PR #3, opened 2016-02-13
Dev/mono support

benhysell:
Updated the mono build for mono 4.2 and cRIO 2015.  Appears to compile/install, but would appreciate a second set of eyes to make sure.


bmouring:
Looks pretty reasonable for 2015 support. I'll pull the changes in and use tags to clarify what's meant for which version

PR #18, opened 2020-09-09
Add CHANGELOG to track notable distro changes

amstewart:
Add a CHANGELOG to the project root, to track notable release changes.

RFC:
* Any objections to tracking changes for nilrt.git, meta-nilrt, and ni/linux in this CHANGELOG; rather than using independent changelogs in the respective repos?
* Any objections to how I have translated the legacy changelog?

Once this PR is merged, I'll make a similar CHANGELOG for the `nilrt/master/dunfell` branch. It will be slightly annoying to keep them both in sync, but it's the price we pay for having divergent mainlines. :/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,221 @@
+# Changelog
+
+This changelog consolidates changes for the [nilrt](https://github.com/ni/nilrt) repo, [meta-nilrt](https://github.com/ni/meta-nilrt) OE layer, and NI's branch of the [linux](https://github.com/ni/linux) kernel. Other OE layer changes are tracked in their respective changelogs.
+
+The NI Linux Real-Time project uses a mainline branching model with product branches for stable releases.
+
+This changelog attempts to conform to the changelog spec on [keepachangelog.org](https://keepachangelog.com/en/1.0.0/).
+
+## [Unreleased]
+Branch: `nilrt/master/sumo`
+
+### NILRT
+
+#### Changed
+- [Upgraded](https://github.com/ni/openembedded-core/commit/a4323b4c193296d499489567b79992c4760fcbc8) curl version to 7.72.0
+- [Upgraded](https://github.com/ni/openembedded-core/commit/ed07362455e469415db3fed805faabcac58bfefb) openssl version to 1.0.2u
+
+#### Deprecated
+- [Drop support](https://github.com/ni/meta-nilrt/commit/1e6453ebca8735de96eaaf3bc931d22998c8dfb3) for VxWorks in curl and openssl
+
+#### Security
+
+- curl fixes
+  - CVE-2018-16839 SASL password overflow via integer overflow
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,182 @@
+# Changelog
+
+This changelog consolidates changes for the [nilrt](https://github.com/ni/nilrt) repo, [meta-nilrt](https://github.com/ni/meta-nilrt) OE layer, and NI's branch of the [linux](https://github.com/ni/linux) kernel. Other OE layer changes are tracked in their respective changelogs.
+
+The NI Linux Real-Time project uses a mainline branching model with product branches for stable releases. The distro itself is versioned according to a yearly release cadence and *is not* semantically versioned.
+
+## [Unreleased]
+Branch: `nilrt/master/sumo`
```
gratian, line 10: Should this say dunfell not sumo?

amstewart: My thoughts were: since we have and will likely continue to have two release mainlines running in parallel, that this branch `master/sumo` would have a CHANGELOG file with the `[Unreleased]` branch being sumo, and the `master/dunfell` ref would have the same, but for dunfell.

When one of the branches goes stable, we can cherry-pick the changelog for that stable version over to the other branch.

What do you think? Is there an easier way to track those changes in parallel?

gratian: Makes sense. Should have payed attention which branch this is going into. Disregard my comment.

CHANGELOG.md
```diff
@@ -0,0 +1,182 @@
+# Changelog
+
+This changelog consolidates changes for the [nilrt](https://github.com/ni/nilrt) repo, [meta-nilrt](https://github.com/ni/meta-nilrt) OE layer, and NI's branch of the [linux](https://github.com/ni/linux) kernel. Other OE layer changes are tracked in their respective changelogs.
+
+The NI Linux Real-Time project uses a mainline branching model with product branches for stable releases. The distro itself is versioned according to a yearly release cadence and *is not* semantically versioned.
```
jeminor: I might suggest just removing the sentence on yearly release, since we may decide to change it more or less often, include bug fixes/patches, etc.

CHANGELOG.md
```diff
@@ -0,0 +1,182 @@
+# Changelog
```
jeminor, line 1: Please add a step to the "Going final" list if you haven't already.

jeminor (changes requested):

CHANGELOG.md
amstewart: I will once this has been accepted.

CHANGELOG.md
amstewart: Fixed.

amstewart:
Patch V2:
* Removed statement about annual releases (jminor's comment above)
* Linked to the changelog spec we're using.
* Added new entries for meta-nilrt changes between when this PR was created and now.

PR #20, opened 2020-09-22
Add a simple pipeline yml to facilitate builds.

jeminor:
Let me know in the build script what all you think we should build.  I just dropped some suggestions, but we need to keep it as minimal as possible so it will build quickly.

Once this is in, I'll create similar pipelines in each of the other sub-repos, I don't see an easier way to do this, but have something that works.

### DIFF
--- a/scripts/azdo/pipeline.yml
+++ b/scripts/azdo/pipeline.yml
@@ -1,0 +1,24 @@
+trigger:
+- nilrt/master/*
+
+resources:
+  containers:
+  - container: rtos-build
+    image: rnd-builds.repos.natinst.com/rtos/nilrt-build:0.3
+    options:
+
+jobs:
+- job: Build_OpenEmbedded
+  container: rtos-build
+  timeoutInMinutes: 3600
+  pool:
+    name: Drivers-NIBuildFarm-RFMIBUILD
+    demands:
+      - docker
+      - Agent.OS -equals Linux
+  steps:
+  - checkout: self
+    submodules: true
+  - script: |
+      source ni-oe-init-build-env
+      ../scripts/buildall.sh
--- a/scripts/buildall.sh
+++ b/scripts/buildall.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+# Used for PR builds, build a variety of things
+export MACHINE=x64
+bitbake nilrt-dkms-image
+bitbake packagegroup-ni-desirable
+


buildall.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Used for PR builds, build a variety of things
+export MACHINE=x64
+bitbake packagegroup-ni-coreimagerepo
```
shruthi-ravi: Is the near-term goal here to build everything(x64) NIOE builds today and have os-common build on top of it? Or are you looking farther out?

shruthi-ravi (commented):
What changes are you making to the submodules in sources/?

buildall.sh
jeminor: My goal is to only use this for PR builds at the moment, since they are difficult to do in the other pipeline (that will do the official exports).  Eventually this might be the thing that really does the build, but it's low overhead to get this working.

jeminor:
> What changes are you making to the submodules in sources/?

I'm not intending on changing anything, I'm not sure why it's showing changes.  Perhaps a merge conflict or something?  I'll try to resolve all of that when I do a test build once we agree on what needs to actually be built :)

buildall.sh
amstewart: If the intention is to check PRs in the fastest, most informative way, then I would recommend just bitbaking the x64 image targets and the desirables packagegroup.

```bash
bitbake nilrt-dkms-image
bitbake packagegroup-ni-desirable
```

That should be enough to trigger bitbake to build nearly all of what goes into the core feeds, and all of the extras packages which we guarantee.

The targets you have identified here are from the nilrt-nxg pipeline, btw. We don't build the coreimagerepo packagegroup and the restore-mode-image in the OneRT pipeline.

amstewart (changes requested):

buildall.sh
jeminor: Fixed in latest version.

shruthi-ravi (approved):
Great work getting this up and running!

amstewart (changes requested):
Could you move the `buildall.sh` script to the `:scripts/` directory, just for cleanliness. Otherwise, looks good to me.

jeminor:
> Could you move the `buildall.sh` script to the `:scripts/` directory, just for cleanliness. Otherwise, looks good to me.

Done.

PR #24, opened 2020-10-09
AzDO handles submodules poorly, force an update on all submodules.

jeminor:


scripts/azdo/pipeline.yml
```diff
@@ -20,5 +20,6 @@ jobs:
   - checkout: self
     submodules: true
   - script: |
+      git submodule update --remote --checkout
```
shruthi-ravi, line 23: Do you need to include the --init option?

shruthi-ravi (approved):
Looks good.

scripts/azdo/pipeline.yml
jeminor: That's a good question, but no (the submodules: true does that much at least).

scripts/azdo/pipeline.yml
```diff
@@ -20,5 +20,6 @@ jobs:
...
       source ni-oe-init-build-env
```
amstewart, line 24: I've pulled the PR which updates this init script with the correct (new) environment variables. Change this source line to [this file](https://github.com/ni/nilrt/blob/nilrt/master/dunfell/nilrt-build-init.env) to populate `NILRT_ROOT`.

amstewart (changes requested):

amstewart:
Is this PR still live, @jeminor ?

PR #31, opened 2020-11-12
buildVM: enable bridged networking in QEMU VMs

amstewart:
Add arguments to the QEMU run-VM scripting, which will enable users to
easily add the VM to an existing network bridge on their host machine.

Also, randomly assign the MAC address of the primary adapter to
discourage IP assignment collisions between multiple VMs on the same subnet.

Randomizing the MAC address means that VMs will have inconsistent IPs
between domain restarts. Users who need a consistent IP should assign
a static MAC address using the new `-a` argument.

----

Calling the script with no arguments will initialize the primary NIC with user-mode networking and a randomly-assigned MAC address. Using `-b $bridge_name` will create a TAP interface and attach it to the bridge with name `$bridge_name`; the MAC address will be random. Using `-a $mac_address` will assign either the user-mode NIC or the bridge with a static MAC address.

### DIFF
--- a/scripts/buildVM-files/runQemuVM.sh
+++ b/scripts/buildVM-files/runQemuVM.sh
@@ -1,23 +1,45 @@
 #!/bin/bash
 set -e
 
-function usage() {
+# Populate the $macaddr variable with a randomly generated MAC address in the
+# 52:54:::: range.
+generate_random_mac() {
+	printf -v macaddr "52:54:%02x:%02x:%02x:%02x" \
+		$(( $RANDOM & 0xff )) $(( $RANDOM & 0xff )) \
+		$(( $RANDOM & 0xff )) $(( $RANDOM & 0xff ))
+}
+
+# Print usage information to stdout.
+# STDOUT: help text
+usage() {
 	cat <<EOF
-$(basename "$0") [-hs] [-c cpu_count] [-m memory] [-- qemu_args [qemu_args]]
+$(basename "$0") [-h] \\
+    [-a mac_address] [-b bridge] [-c cpu_count] [-m memory] [-s] \\
+    [-- qemu_args [qemu_args]]
 
 Opts:
-	-h    : Display this help and exit.
-	-s    : Start VM in snapshot mode (changes not saved to disk)
+    -h    : Display this help and exit.
+    -s    : Start VM in snapshot mode (changes not saved to disk)
 
 Args:
-	cpu_count : the number of CPUs which will be available to the VM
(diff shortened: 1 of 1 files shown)


amstewart:
@shruthi-ravi @ni/rtos

scripts/buildVM-files/runQemuVM.sh
```diff
@@ -1,23 +1,45 @@
 #!/bin/bash
 set -e
 
-function usage() {
+# Populate the $macaddr variable with a randomly generated MAC address in the
+# 52:54:::: range.
+generate_random_mac() {
+	printf -v macaddr "52:54:%02x:%02x:%02x:%02x" \
+		$(( $RANDOM & 0xff )) $(( $RANDOM & 0xff )) \
```
shruthi-ravi, line 8: Do you need to seed RANDOM here before using it?

scripts/buildVM-files/runQemuVM.sh
```diff
@@ -62,4 +94,6 @@ qemu-system-x86_64 \
 	-drive if=pflash,format=raw,readonly,file="$SCRIPT_DIR/OVMF/OVMF_CODE.fd" \
 	-drive if=pflash,format=raw,file="$SCRIPT_DIR/OVMF/OVMF_VARS.fd" \
 	-drive file="$SCRIPT_DIR/${VM_NAME}.qcow2",index=0,media=disk \
+	-device e1000,netdev=nilrt_net0,mac=$macaddr \
```
shruthi-ravi, line 97: Should this e1000 driver name also be configurable from the command line?

scripts/buildVM-files/runQemuVM.sh
amstewart: The default seed for `$RANDOM` seems sufficient for what I'm using it for here. I stole it from a [helper script](https://github.com/ni/nilrt/blob/d77f1219b48ab71402c329eabd93e924679b2f64/scripts/tests/vpn-test-files/start_qemu_bridge.sh#L17) for the provisioning tests - which has a history of creating sufficiently random MAC addresses for its VMs.

scripts/buildVM-files/runQemuVM.sh
amstewart: I think the 99% use-case for this shell script in particular, is by an interactive user who is trying to start a VM quickly. There might be some cases where we would want to emulate a particular PHY driver for ATS purposes, but I don't see an argument in this script being used very frequently.

shruthi-ravi (approved):

PR #38, opened 2021-02-09
scripts/azdo: add AZDO component conf files

amstewart:
Add an auto.conf file, containing bitbake configuration settings which
(a) are common to all NI-internal AZDO bitbake components and (b) have
static values.

Add an ni-org.conf file, containing bitbake configuration settings which
are necessary to produce faster, more-reproducible builds, but which are
insane for an external (non-NI) builder.

----

### Testing

* Rebuilt `rtos_oe_feeds` on my dev machine and verified that the variables I moved have their correct values and that all makefile targets complete.

### Associated PRs
* meta-nilrt [PR 125](https://github.com/ni/meta-nilrt/pull/125) for the nilrt.git angle on this change.
* [AZDO](https://dev.azure.com/ni/DevCentral/_git/ni-central/pullrequest/123493)

### DIFF
--- a/nilrt-build-init.env
+++ b/nilrt-build-init.env
@@ -2,6 +2,21 @@
 
 SCRIPT_ROOT=$(realpath $(dirname ${BASH_SOURCE:-${0}}))
 
+enable_ni_org_conf=false
+positionals=()
+
+while [ $# -ge 1 ]; do case "$1" in
+	-o|--org)
+		enable_ni_org_conf=true
+		shift
+		;;
+	*)
+		positionals+=($1)
+		shift
+		;;
+esac; done
+
+
 BITBAKEDIR=${BITBAKEDIR:-${SCRIPT_ROOT}/sources/bitbake}
 NILRT_ROOT="${NILRT_ROOT:-${SCRIPT_ROOT}}"
 
@@ -16,7 +31,15 @@ export NILRT_ROOT
 
 # Call OE-upstream's build env initialization script, which will create a build
 # workspace called either "${1:-build}/" and `cd` into it.
-. $PWD/sources/openembedded-core/oe-init-build-env
+cd ${SCRIPT_ROOT}
+. ./sources/openembedded-core/oe-init-build-env ${positionals[@]}
+
+if $enable_ni_org_conf; then
+	if [ ! -e conf/site.conf ]; then
+		echo "Adding NI org.conf as conf/site.conf..."
+		cp ${SCRIPT_ROOT}/scripts/azdo/conf/ni-org.conf ./conf/site.conf
+	fi
+fi
 
 # Add a marker to the prompt based on whether or not bitbake is in the
 # environment.
--- a/scripts/azdo/conf/auto.conf
+++ b/scripts/azdo/conf/auto.conf
@@ -1,0 +1,12 @@
+# Include buildhistory as it is used by toaster as well as generally
+# useful in detecting changes in builds/configurations
+INHERIT += "buildhistory"
+BUILDHISTORY_COMMIT = "1"
+
+# Generates a "cve/cve.log" in every recipe's work dir.
+# https://wiki.yoctoproject.org/wiki/How_do_I#Q:_How_do_I_get_a_list_of_CVEs_patched.3F
+#INHERIT += "cve-check"
+
+# The buildstats class records performance statistics about each task executed
+# during the build (e.g. elapsed time, CPU usage, and I/O usage).
+USER_CLASSES += "buildstats"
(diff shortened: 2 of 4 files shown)


amstewart:
@ni/rtos

scripts/azdo/conf/auto.conf
```diff
@@ -0,0 +1,14 @@
+include conf/ni-org.conf
```
shruthi-ravi: I'm not convinced that this needs to be a separate conf file from ni-org.conf. This too will not be used by anyone outside NI, right? 
I'm also concerned that for a developer build outside azdo, which is a common-enough case, the configuration in these two files will not be included in the build by default. What are your thoughts about that?

amstewart: I broke out configuration into the auto.conf which enabled unnecessary build artifacts (buildhistory, cve logs, et c.). My thinking was that NI devs can use the auto.conf, but not waste time making those extraneous artifacts, unless they want to explicitly add them to their local.conf.

> This too will not be used by anyone outside NI, right?

Correct.

>I'm also concerned that for a developer build outside azdo, which is a common-enough case, the configuration in these two files will not be included in the build by default. What are your thoughts about that?

I suppose we can add a switch to [`nilrt-build-init.env`](https://github.com/ni/nilrt/blob/nilrt/master/dunfell/nilrt-build-init.env) which adds the ni-org.conf to your build environment automatically. Would that suffice?

amstewart:
PATCH V2:
* added a switch (`-o` | `--org`) to `nilrt-build-init.env`, which will add the `ni-org.conf` snippet as `conf/site.conf`, for use by developers. External builders can still just source the init file with no arguments, and it will work in the same way as before.

nilrt-build-init.env
```diff
@@ -2,6 +2,21 @@
 
 SCRIPT_ROOT=$(realpath $(dirname ${BASH_SOURCE:-${0}}))
 
+enable_ni_org_conf=false
+positionals=()
+
+while [ $# -ge 1 ]; do case "$1" in
+	-o|--org)
+		enable_ni_org_conf=true
```
billpittman, line 10: I'm torn, the selfish side of me wants our use case to be the "no parameters - just add in the NI stuff".  However I also recognize that users will "always" fail if they just invoke the script without parameters.  So I understand why you did the logic the way you did.

shruthi-ravi, line 10: I found the commit wording confusing and my bash knowledge is minimal, so let me confirm my understanding here. An NI user now has to invoke this script with the -o parameter `. nilrt-build-init.env -o` to copy the ni-org.conf file to the build directory. Correct?

scripts/azdo/conf/auto.conf
shruthi-ravi: > I suppose we can add a switch to nilrt-build-init.env which adds the ni-org.conf to your build environment automatically. Would that suffice?

I guess there aren't better alternatives. Can you add a note to the README recommending how to invoke the script for NI developers? And announce this change to the team.

nilrt-build-init.env
billpittman: shruthi - yes that's my read, and I'm not completely happy with that solution.

@amstewart Could we possibly make the script a bit smarter and check the user's email address and if its an @ni.com email, initialize enable_ni_org_conf=true ?  'git config user.email |grep "\@ni.com$"'  - or some such ??  That way the script would basically do the right thing most of the time (assuming we all have (or set) our email addresses with git).

amstewart: I figure NI developers are smart enough to know how to alias or wrap the script in their workspace to automatically use `--org`.

amstewart: I don't like the idea of the script trying to be smart like that. Having an NI email address configured doesn't _mean_ that you are connected to the corporate network. I disconnect from the VPN all the time and still want to do bitbake things, while my repo still has an NI email attached to it.

I've considered a few different heuristics for determining if you're connected to the corporate network. But none of them really get around the fundamental requirement that VPN builders understand that these config settings are a thing, and that they will only work while connected to the VPN.

I think its an improvement that they are broken out specifically from the rest of the local.conf. And I like the implication of the `-o` switch on the initscript; it implies "Yes, I understand the implications of adding these config settings and want to be responsible for turning them off again, if I need to."

As I said above, you can just wrap the script on your local machine with a one-line source file that provides the `-o` argument, or edit the initscript locally.

shruthi-ravi (approved):
lgtm as long as changes are communicated to the team so that noone is caught by surprise.

nilrt-build-init.env
billpittman: Too true, you've got a couple good points there.  I guess I'll have to write a script or remember something ;)

billpittman (approved):

amstewart:
### Patch V3
* Updated the `README.md` to advertise the new `--org` argument and added a section telling NI developers to use it.

shruthi-ravi (approved):

amstewart:
### Patch V4
* added a fixup to the `auto.conf` script to remove the `include conf/ni-org.conf` line. I decided to install the `ni-org.conf` script as `conf/site.conf` in the AZDO components, which is automatically included, so the includes line is unnecessary.

PR #41, opened 2021-03-04
Get the provisioning ISO working

amstewart:
This patchset...
1. switches the default provisioning scheme for the automated VM creation scripts to `efi-ab` - which is the correct scheme for RAUC targets. This change is necessary to get the [provisioning ptest](https://github.com/ni/meta-nilrt/blob/nilrt/master%2Fdunfell/recipes-core/initrdscripts/files/ptest/test_ni_provisioning.sh) working. 
2. deprecates the non-QEMU virtual machine flavors from the buildVM script. They have not been officially supported for a long time and they do not support the `efi-ab` boot scheme. There will be a concomitant PR to the AZDO components to fixup the script arguments in the makefiles.
3. adds a new `test_provisioning` script which uses QEMU and tcl-expect to instrument the `init-restore-mode-ptest` test. The AZDO PR will add the new subcomponent which makes use of this script to block the AZDO build when the ISO has created a nonfunctional VM.

## Testing

* Ran the new provisioning test many times both manually and through nibuild.

@ni/rtos @jpautler


scripts/buildVM-files/ni_provisioning.answers
```diff
@@ -3,8 +3,6 @@
 # Full automated installation; yes on all options + poweroff on finish
 PROVISION_UNSUPPORTED_TARGET="y"
 PROVISION_REPARTITION_TARGET="y"
-PROVISION_RESTORE_MINIMAL_IMAGE="y"
```
shruthi-ravi, line 6: What's this change about?

scripts/tests/provisioning/test.expect
```diff
@@ -0,0 +1,65 @@
...
+# run ptests
+expect -re "$USER@.*# $" {
+	send "ptest-runner -l\r"
+	send "ptest-runner\r"
+}
+
+# shutdown
+expect {
+	-re "$USER@.*# $" { send -- "halt\r" }
+	timeout { exit 1 }
```
shruthi-ravi: Can you add some text to describe what timed out? The ptest-runner hung?

scripts/tests/provisioning/test_provisioning.sh
```diff
@@ -0,0 +1,163 @@
...
+	echo "ERROR: " $@ >&2
+	exit 1
+}
+
+
+## ARGUMENT PARSING AND ENVIRONMENT
+
+usage() {
+	local exit_code=${1:-2}
+	test $exit_code -eq 0 && exec 1>&2
```
shruthi-ravi: Shouldn't this be -neq? Redirect stdout to stderr when exit_code neq 0. Have I misunderstood?

shruthi-ravi (approved):
Minor. Nice work!

scripts/buildVM-files/ni_provisioning.answers
```diff
@@ -3,8 +3,6 @@
 # Full automated installation; yes on all options + poweroff on finish
 PROVISION_UNSUPPORTED_TARGET="y"
```
jpautler: I assume you removed PROVISION_RESTORE_MINIMAL_IMAGE because it is never referenced by the provisioning scripts (was removed at some point in the past maybe?). If so, can you remove PROVISION_UNSUPPORTED_TARGET as well since it seems to never be referenced either.

scripts/buildVM-files/ni_provisioning.answers
amstewart: As far as I can tell, `PROVISION_RESTORE_MINIMAL_IMAGE` is not used by any of our scripts and - I would assume - refers to the nxg "minimal" images which we have deprecated. I removed the setting for cleanliness.

scripts/tests/provisioning/test.expect
amstewart: Sure; will fix.

scripts/tests/provisioning/test_provisioning.sh
amstewart: Yep. Good catch. Will fix.

scripts/tests/provisioning/test.expect
```diff
@@ -0,0 +1,65 @@
...
+}
+
+expect EOF
```
jpautler: I'm not sure, but should we have a timeout associated with the run-vm script exiting within a reasonable amount of time? For example, if the target VM gets stuck during shutdown should we consider that an error?

This test will probably become a reference for future tests, so if we think that would be useful maybe we want to illustrate it here? If not, feel free to drop this suggestion.

jpautler (approved):

scripts/buildVM-files/ni_provisioning.answers
amstewart: Correct. I'll see if I can remove the UNSUPPORTED_TARGET var as well.

shruthi-ravi: Can you update the commit message to reference these changes while you're at it? :)

scripts/tests/provisioning/test.expect
amstewart: The shutdown command will follow the default timeout for the expect script, which is set at 90 seconds. If it takes longer than that, I will print an `TIMEOUT` message and the entire expect process (and VM) will exit with code 1. That will trigger the `set -e` in the `test_provisioning` script and propagate the error to the subcomponent. ie. A hang during shutdown should cause the test to fail sanely.

amstewart:
## Patch V2
* Removed the `PROVISION_UNSUPPORTED_TARGET` option from the answers file, because it is no longer needed.
* Add more version TIMEOUT messages to `test.expect`.
* Fixup the `usage()` output redirection logic in `test_provisioning.sh`.

shruthi-ravi (approved):

scripts/tests/provisioning/test.expect
jpautler: I see. I think I was thrown off by the fact that the expect call just above this had an explicit timeout case. Does that explicit timeout case do anything different than if you had just left it off and relied on the default timeout behavior?

amstewart: See my latest patch-rev.

billpittman (approved):

scripts/tests/provisioning/test.expect
```diff
@@ -34,20 +34,22 @@ expect {
...
+		foreach feed_uri $ipk_feeds {
+			send -- "echo src test-feed.$feed_index http://$vm_host_ip:8990/$feed_uri >>/etc/opkg/testing.conf\r"
+			set feed_index [expr $feed_index + 1]
+		}
+		send "cat /etc/opkg/testing.conf\r"
+		send "opkg update\r"
+		send "opkg install init-restore-mode-ptest\r"
+	}
+	timeout {
+		send_user "TIMEOUT waiting for opkg to configure and ptests to install.\r"
```
jpautler: Isn't this expect clause waiting for login to complete so it can **send** the opkg commands rather than waiting **for** the opkg commands to complete? I think this same issue applies to each timeout clause below.

scripts/tests/provisioning/test.expect
jpautler: If I understand the behavior properly, I think the **expect eof** still needs a timeout clause if we want to return an error code and print a message.

scripts/tests/provisioning/test.expect
```diff
@@ -0,0 +1,65 @@
...
+spawn [lindex $argv 0] -s -- -device e1000,netdev=n1 -netdev user,id=n1,ipv6=off,net=10.0.2.0/24,host=$vm_host_ip
+
+# getty
+# run-postinsts on a non-accelerated VM might take ~300 seconds, so increase
+# the timeout to provide plenty of time.
+set timeout 900
+expect {
+	"login: " { send "root\r" }
+	timeout {
+		send_user "TIMEOUT while waiting for getty."
```
jpautler: Should this line get a \r as well?

scripts/tests/provisioning/test.expect
amstewart: No. The `timeout` clause applies to the time it takes to execute all actions within the parent `expect` clause, not the time between this clause and the previous.

When I artificially cause the `configure opkg` clause to timeout, for example, you get this:
```bash
...
NI-Unknown-52542d724e76 login: root

TIMEOUT waiting for opkg to configure and ptests to install.
INFO: Feed server (2320) terminated.
removed '/tmp/test_provisioning.nDsHD/feed.0'
removed '/tmp/test_provisioning.nDsHD/feed.1'
removed '/tmp/test_provisioning.nDsHD/feed.2'
/home/usr0/projects/ni-central/src/rtos/rtosimagesd/rtos_nilinuxrt/source/nilrt/scripts/tests/provisioning/test_provisioning.sh: line 110:  2320 Terminated              python3 -m http.server $IPK_FEED_PORT  (wd: /tmp/test_provisioning.nDsHD)
make[2]: *** [objects/test_provisioning/linuxU/x64/gcc-4.7-oe/release/ni-linux-rt-x64-qemu/provisioning.log] Error 1

```

scripts/tests/provisioning/test.expect
amstewart: Yeah. Also, in the process of triple checking my understanding to your comment above, I have figured out a better way to write these `send_user` calls. Will fix in V4.

amstewart:
## Patch V4

Using the newline character, instead of the carriage return, is better
for `send_user` in a GNU make context. When using \n before and after
the message, MAKE will avoid its tendancy to drop the first character of
the message output and also to overwrite the message line with the next
stderr output.

Prevously, messages would sometimes look like this:
```
IMEOUT waiting for opkg to configure and ptests to install.
INFO: Feed server (2473) terminated.
```

or this
```
INFO: Feed server (2473) terminated. and ptests to install.
```

Now they are printed in their completion on their own line.

amstewart:
## Patch V5
* @jpautler was correct about the timeout locations. My original verification was incorrect and the messages should have been printing information about why the *previous* clause failed. I moved the messages into their correct locations.
* Used the `$USER` expect variable for getty login, as I had originally intended.

### Testing

* Used an intentionally-wrong username for getty login to confirm that the timeout messages are speaking to the correct failures.
* Re-verified that the test still passes with these changes (when using the correct getty username).

scripts/tests/provisioning/test.expect
amstewart: ^ This is wrong. See Patch V5 below.

scripts/tests/provisioning/test.expect
```diff
@@ -25,7 +25,7 @@ spawn [lindex $argv 0] -s -- -device e1000,netdev=n1 -netdev user,id=n1,ipv6=off
 # the timeout to provide plenty of time.
 set timeout 900
 expect {
-	"login: " { send "root\r" }
+	"login: " { send "$USER\r" }
 	timeout {
 		send_user "\nTIMEOUT while waiting for getty.\n"
```
shruthi-ravi, line 30: > the messages should have been printing information about why the previous clause failed...

If that's the case, what does this timeout apply to?

scripts/tests/provisioning/test.expect
```diff
@@ -47,7 +47,7 @@ expect {
 		send "opkg install init-restore-mode-ptest\r"
 	}
 	timeout {
-		send_user "\nTIMEOUT waiting for opkg to configure and ptests to install.\n"
+		send_user "\nTIMEOUT waiting for shell.\n"
```
shruthi-ravi, line 50: And do you need a 90s timeout for the shell to appear?

scripts/tests/provisioning/test.expect
shruthi-ravi: Nvm, I misunderstood your description.

scripts/tests/provisioning/test.expect
amstewart: You shouldn't *need* one. But that's how I have it written, yes.

gratian (approved):

jpautler (approved):

PR #43, opened 2021-03-11
create_safemode: Provision a efi-ab VM to safemode

chaitu236:
@ni/rtos 

There will be a corresponding azdo PR

### DIFF
--- a/scripts/tests/provisioning/migrate_vm_to_safemode.expect
+++ b/scripts/tests/provisioning/migrate_vm_to_safemode.expect
@@ -1,0 +1,127 @@
+#!/usr/bin/expect -f
+#
+# Start a NILRT efi-ab VM, install grub gateway IPK and migrate to safemode.
+# This script assumes that there is a running HTTP server at the ipv4
+# $vm_host_ip, on port 8990.
+#
+# Expected arguments:
+#   1) path to the run-VM script
+#   2) name of the log file
+#   3+) ipk feed paths (optional)
+log_file -noappend [lindex $argv 1]
+
+set ipk_feeds [lrange $argv 2 end]
+# The SLIRP networking static ipv4 to use for the VM host (as viewed from the
+# guest). Can be any ip on the 10.0.2.0/24 subnet. 10.0.2.2 is the QEMU
+# default, if none is specified.
+set vm_host_ip "10.0.2.2"
+# Set the login username
+set USER root
+
+# boot the VM with the run-vm script
+spawn [lindex $argv 0] -- -device e1000,netdev=n1 -netdev user,id=n1,ipv6=off,net=10.0.2.0/24,host=$vm_host_ip
+
+# getty
+# run-postinsts on a non-accelerated VM might take ~300 seconds, so increase
+# the timeout to provide plenty of time.
+set timeout 900
+expect {
+	"login: " { send "$USER\r" }
+	timeout {
+		send_user "\nTIMEOUT while waiting for getty.\n"
+		exit 1
+	}
+}
+
+set timeout 90
+# configure opkg
+expect {
+	-re "$USER@.*# $" {
+		send "touch /etc/opkg/testing.conf\r"
+		set feed_index 0
+		foreach feed_uri $ipk_feeds {
+			send -- "echo src test-feed.$feed_index http://$vm_host_ip:8990/$feed_uri \[trusted=yes\] >>/etc/opkg/testing.conf\r"
+			set feed_index [expr $feed_index + 1]
+		}
+		send "cat /etc/opkg/testing.conf\r"
+		send "opkg update\r"
+		send "opkg install dist-nilrt-grub-gateway\r"
+	}
+	timeout {
+		send_user "\nTIMEOUT waiting for shell.\n"
+		exit 1
+	}
+}
+
+# start provisioning
+expect {
+	-re "$USER@.*# $" {
+		send "/usr/share/nilrt/nilrt-install\r"
+	}
+	timeout {
+		send_user "\nTIMEOUT waiting for opkg to configure and ptests to install.\n"
+		exit 1
+	}
+}
+
+# reboot
+expect {
+	-re "$USER@.*# $" { send -- "reboot\r" }
+	timeout {
+		send_user "\nTIMEOUT waiting for nilrt-install script to complete.\n"
+		exit 1
+	}
+}
+
+expect {
+	"$USER@recovery:/#" {}
+	timeout {
+		send_user "\nTIMEOUT while waiting for system to start provisioning.\n"
+		exit 1
+	}
+}
+
+# Wait for shell after provisioning is complete
+set timeout 900
+expect {
+	"NI Linux Real-Time (safe mode on" {}
+	timeout {
+		send_user "\nTIMEOUT while waiting for system to boot after provisioning.\n"
+		exit 1
+	}
+}
+
+set USER admin
+
+set timeout 90
+expect {
+	"login: " { send "$USER\r" }
+	timeout {
+		send_user "\nTIMEOUT while waiting for getty.\n"
+		exit 1
+	}
+}
+
+expect {
+	"Password: " { send "\r" }
+	timeout {
+		send_user "\nTIMEOUT while waiting for getty.\n"
+		exit 1
+	}
+}
+
+# shutdown
+expect {
+	-re "$USER@.*# $" { send -- "halt\r" }
+	timeout {
+		send_user "\nTIMEOUT waiting for login.\n"
+		exit 1
+	}
+}
+
+expect {
+	EOF
+	timeout {
+		send_user "\nTIMEOUT waiting for VM to shut down.\n"
+	}
+}
--- a/scripts/tests/provisioning/test_provisioning.sh
+++ b/scripts/tests/provisioning/test_provisioning.sh
@@ -7,154 +7,17 @@
 #  * configures opkg with a locally-hosted ipk feed.
 #  * installs and runs the provisioning ISO's ptest package.
 #  * evaluates the ptest results.
-set -euo pipefail
 
-readonly SCRIPT_ROOT=$(dirname "$BASH_SOURCE")
-
-
-## UTILITY FUNCTIONS
-
-error_and_die() {
-	echo "ERROR: " $@ >&2
-	exit 1
-}
-
-
-## ARGUMENT PARSING AND ENVIRONMENT
-
-usage() {
-	local exit_code=${1:-2}
-	test $exit_code -eq 0 || exec 1>&2
-	cat <<EOF
-$(basename $BASH_SOURCE) -h|--help
-  Print this usage information and exit.
-$(basename $BASH_SOURCE) [-i ipk_feed_dir] nilrt_vm_path
-  Start the NILRT VM and run the provisioning tests.
-
-Options:
--i|--ipk-feed  Path to a local IPK feed, which will be added to the VM's opkg
-               configuration.
-
-Arguments:
-nilrt_vm_path  Path to an already-provisioned NILRT VM directory. (See the
-               buildVM.sh script in the NILRT.git source.)
-EOF
-	exit $exit_code
-}
-
-readonly IPK_FEED_PORT=8990
-readonly EXPECT=/usr/bin/expect
-
-# Check environment requirements
-test -x "$EXPECT" || error_and_die "$EXPECT not found. This script requires the tcl-expect package to instrument test operations."
-
-VERBOSE=${VERBOSE:-}
-ipk_feeds=()
-positionals=()
-
-while test -n "${1:-}"; do case "$1" in
-	-h|--help)
-		usage 0
-		;;
-	-i|--ipk-feed)
-		ipk_feeds+=("$2")
-		shift
-		shift
-		;;
-	-*|--*)
-		echo "Invalid or unrecognized option: $1"
-		usage 2
-		;;
-	*)
-		positionals+=("$1")
-		shift
-		;;
-esac; done
-
-if [ ${#positionals} -lt 1 ]; then
-	usage 2
-fi
-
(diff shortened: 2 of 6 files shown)


scripts/tests/provisioning/create_safemode.expect
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/expect -f
```
gratian: Can we rename this to something more explicit since this is not really creating a safemode? It will make it easier to remember what this does in the future. Something like 'test_efi_ab_to_safemode.expect' or 'test_rauc_to_safemode.expect'. FWIW I also don't like the existing generically named  'test.expect'. I might try to rename that in the PR that will go with my changes.

scripts/tests/provisioning/create_safemode.expect
```diff
@@ -0,0 +1,112 @@
...
+#
+# Start a NILRT VM and instrument installing and running the provisioning
+# ptests. This script assumes that there is a running HTTP server at the ipv4
+# $vm_host_ip, on port 8990.
+# 
+# Expected arguments:
+#   1) path to the run-VM script
+#   2+) ipk feed paths (optional)
+log_file -noappend [lindex $argv 1]
```
gratian: Does the comment above need updating to account for the 'log_file' argument?

shruthi-ravi:
@chaitu236 can you describe why you need to make these changes?

scripts/tests/provisioning/create_safemode.expect
chaitu236: This file is intended for just creating safemode VM not testing. There is another expect file for actually running the ptest.
I was planning to run the create_safemode.expect in azdo subcomponent that builds the gateway ipk, and run the actual test expect file in another subcomponent that does the testing. This way your test subcomponent can consume the safemode VM directly.

chaitu236:
> @chaitu236 can you describe why you need to make these changes?

create_safemode.expect is intended to be run in the azdo subcomponent that builds rauc->safemode gateway ipk.
The expect file creates a safemode VM (no tests)

The test_safemode.expect file that I just pushed is intended to be run by another azdo subcomponent that runs the ptests.
The safemode->rauc tests that you/Gratian are working on can consume the VM from first subcomponent.

scripts/tests/provisioning/configure_vm.sh
```diff
@@ -0,0 +1,151 @@
...
+## ARGUMENT PARSING AND ENVIRONMENT
+
+usage() {
+	local exit_code=${1:-2}
+	test $exit_code -eq 0 || exec 1>&2
+	cat <<EOF
+$(basename $BASH_SOURCE) -h|--help
+  Print this usage information and exit.
+$(basename $BASH_SOURCE) [-i ipk_feed_dir] nilrt_vm_path
+  Start the NILRT VM and run the provisioning tests.
```
shruthi-ravi: Can you update the comment to the current usage?

scripts/tests/provisioning/test_provisioning.sh
```diff
@@ -7,154 +7,16 @@
...
-ipk_feeds_names=$(find ${feed_server_pubdir} -maxdepth 1 -type l -printf "%f\n")
-
-# Let tcl-expect drive the process of running the test, using the QEMU serial console.
-# It should record all stdout to ./provisioning.log in the VM dir.
-"${SCRIPT_ROOT}/test.expect" \
-	${vm_run_script} \
-	${ipk_feeds_names}
-
-feed_server_teardown
+source configure_vm.sh
```
shruthi-ravi: Don't you have to provide an expect file here and other arguments?

scripts/tests/provisioning/configure_vm.sh
```diff
@@ -0,0 +1,151 @@
...
+
+# Find the VM "run" helper script in the machine directory
+vm_run_script=$(find ./ -name 'run-*.sh' -exec realpath {} \;)
+test -n "${vm_run_script}" || error_and_die "Could not find VM run script in \"${nilrt_vm_path}\""
+
+# Names like: feed.0, feed.1, et c.
+ipk_feeds_names=$(find ${feed_server_pubdir} -maxdepth 1 -type l -printf "%f\n")
+popd
+
+# Let tcl-expect drive the process of running the test, using the QEMU serial console.
```
shruthi-ravi: I think the comments should be reworked so they don't focus on testing as much. Testing depends on the expect script passed in?

scripts/tests/provisioning/create_safemode.expect
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/expect -f
+#
+# Start a NILRT VM and instrument installing and running the provisioning
```
shruthi-ravi: Update comment to not include ptests?

scripts/tests/provisioning/test_safemode.expect
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/expect -f
+#
+# Start a safemode NILRT VM and instrument installing and running the provisioning
```
shruthi-ravi: Should this be referring to grub gateway ptests and not provisioning ptests? Are they the same? Or generically leave it at ptests?

scripts/tests/provisioning/test_safemode.expect
```diff
@@ -0,0 +1,89 @@
...
+# ptests. This script assumes that there is a running HTTP server at the ipv4
+# $vm_host_ip, on port 8990.
+#
+# Expected arguments:
+#   1) path to the run-VM script
+#   2+) ipk feed paths (optional)
+log_file -noappend [lindex $argv 1]
```
shruthi-ravi: Add to expected args?

scripts/tests/provisioning/test_safemode.expect
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/expect -f
```
shruthi-ravi: Can you rename the test to something more descriptive like, test_migration_to_safemode?

scripts/tests/provisioning/configure_vm.sh
```diff
@@ -0,0 +1,151 @@
+#!/bin/bash
```
shruthi-ravi: I feel like this script does much more than configuring a VM, and would benefit from a rename.

shruthi-ravi (commented):
Overall, I think comments and filenames could be reworked due to the design change, for readability.

scripts/tests/provisioning/create_safemode.expect
gratian: OK that makes sense but it's still not creating safemode. Maybe create_safemode_vm.expect then? Or more accurately provision_safemode_vm.expect or migrate_vm_to_safemode.expect.

scripts/tests/provisioning/create_safemode.expect
```diff
@@ -0,0 +1,112 @@
...
+	}
+}
+
+# start provisioning
+expect {
+	-re "$USER@.*# $" {
+		send "/usr/share/nilrt/nilrt-install\r"
+	}
+	timeout {
+		send_user "\nTIMEOUT waiting for nilrt-install to complete.\n"
```
billpittman: For expect scripts, aren't the timeouts for the previous command ?  So if this timeout triggers, it was actually the series of opkg commands that timed out, not the nilrt-install script.

scripts/tests/provisioning/create_safemode.expect
```diff
@@ -0,0 +1,112 @@
...
+
+# reboot
+expect {
+	-re "$USER@.*# $" { send -- "reboot\r" }
+	timeout {
+		send_user "\nTIMEOUT waiting for reboot command to complete.\n"
+		exit 1
+	}
+}
+
```
billpittman: In the interest of not timing out, and getting one more data point, I think you should add an expect for the "recovery" prompt.  Just to acknowledge that we successfully booted into the recovery partition and that provisioning started.

billpittman (changes requested):

scripts/tests/provisioning/configure_vm.sh
amstewart: Maybe `start-vm-using-expect.sh`?

scripts/tests/provisioning/create_safemode.expect
amstewart: ^ Bill is correct. Maybe you need to merge the [latest version](https://github.com/ni/nilrt/blob/b126ec18aa7788feadcb649a2e61646efa60aa99/scripts/tests/provisioning/test.expect) of my `test.expect` file?

scripts/tests/provisioning/test_provisioning.sh
```diff
@@ -7,154 +7,16 @@
...
 
 # Quickly parse the test session log for pass/fail status
 # GREP will return code...
 #   0, if the ptest PASS line is present
 #   1, if the line is not present (or is FAIL: or SKIP:)
-#   2, if the provisioning.log does not exist for whatever reason
+#   2, if the log file does not exist for whatever reason
 echo "INFO: Parsing the test log..."
-if grep -E '^PASS: ni_provisioning\W' ./provisioning.log; then
+if grep -E '^PASS: ni_provisioning\W' ./${log_file}; then
```
amstewart, line 20: Where is `log_file` defined?

amstewart (changes requested):

scripts/tests/provisioning/test_provisioning.sh
gratian: FYI when this script is called from the rtosimagesd component this fails with:
.../ni-central/src/rtos/rtosimagesd/rtos_nilinuxrt/source/nilrt/scripts/tests/provisioning/test_provisioning.sh: line 11: configure_vm.sh: No such file or directory

I think you need an absolute path or make it relative to the current folder.

scripts/tests/provisioning/configure_vm.sh
```diff
@@ -0,0 +1,151 @@
...
+		shift
+		;;
+esac; done
+
+if [ ${#positionals} -lt 1 ]; then
+	usage 2
+fi
+
+nilrt_vm_path=${positionals[0]}
+expect_file=${positionals[1]}
```
gratian: FYI when called from rtosimagesd/rtos_nilinuxrt component this line errors out with:
ni-central/src/rtos/rtosimagesd/rtos_nilinuxrt/source/nilrt/scripts/tests/provisioning/configure_vm.sh: line 77: positionals[1]: unbound variable
make[1]: *** [objects/test_provisioning/linuxU/x64/gcc-4.7-oe/release/ni-linux-rt-x64-qemu/provisioning.log] Error 1

It is possible I have a dirty build since I'm hand merging dev branches but it might be something to check.

scripts/tests/provisioning/configure_vm.sh
```diff
@@ -0,0 +1,151 @@
...
+		echo "Invalid or unrecognized option: $1"
+		usage 2
+		;;
+	*)
+		positionals+=("$1")
+		shift
+		;;
+esac; done
+
+if [ ${#positionals} -lt 1 ]; then
```
gratian: Should this be '-lt 2' now that the VM path and expect file are positional params?

scripts/tests/provisioning/configure_vm.sh
gratian: Ah, never mind the azdo component was not updated yet to account for the new parameter.

scripts/tests/provisioning/test_provisioning.sh
chaitu236: That isn't necessary since we're sourcing the file.

scripts/tests/provisioning/test_safemode.expect
chaitu236: They should be the same right? I think keeping "provisioning ptests" which implies we're testing provisioning is more helpful than just "ptests".

scripts/tests/provisioning/test_provisioning.sh
shruthi-ravi: Sorry, I don't understand. How are the tests in `test.expect` run in this case?

scripts/tests/provisioning/test_provisioning.sh
chaitu236: Its defined in configure_vm.sh which is sourced from this file.

scripts/tests/provisioning/test_provisioning.sh
chaitu236: Since configure_vm.sh is sourced in this file, it is equivalent to just including everything from configure_vm.sh in place of the `source configure_vm.sh`.
Arguments to test_provisioning.sh will still be accessible to code in configure_vm.sh

chaitu236: I hadn't completely understood Shruthi's question. The original test_provisioning.sh invoked test.expect. I am now changing it so that the expect file is passed in as parameter so that both safemode and rauc ptests can be run by test_provisioning.sh.
The intent is to pass the expect file in azdo subcomponent mak file.

chaitu236:
Update 3: Addressed review feedback

amstewart (approved):

scripts/tests/provisioning/migrate_vm_to_safemode.expect
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+# start provisioning
+expect {
+	-re "$USER@.*# $" {
+		send "/usr/share/nilrt/nilrt-install\r"
+	}
+	timeout {
+		send_user "\nTIMEOUT waiting for opkg to configure and ptests to install.\n"
```
shruthi-ravi, line 62: Update this old timeout message. Tests aren't installed here?

gratian (approved):

shruthi-ravi (approved):
Final nit. Looks good.

billpittman (approved):

PR #44, opened 2021-03-17
Many updates to the buildVM script

amstewart:
While debugging the issues in [meta-nilrt PR 142](https://github.com/ni/meta-nilrt/pull/142), I made several quality of life improvements to the buildVM script.

1. cleaned up the `usage()` function to be more correct and elegant.
2. allowed user-interaction with the provisioning script.
3. make the script verbose, by default.
4. stop forcing the user to set `MACHINE=x64` unnecessarily.
5. improve KVM detection @gratian 
    * I hope the improved detection will fix the [AZDO build failure here](https://dev.azure.com/ni/DevCentral/_build/results?buildId=773463&view=logs&j=44e70749-ceba-5737-fad1-33069546191c&t=a2065b91-4596-5839-9588-2e1a5df2fc02&l=23108).

---

## TESTING
* I've been using this new script for a while.
* I have not tested the new KVM detection on the AZDO build machines yet. I will not accept this PR until that part is validated.

**NOTICE**
This PR includes some breaking API changes for now the `buildVM.sh` script is called (it deprecates the `-v` and `-q` options). The AZDO pipeline will break without concomitant changes to the Makefiles. I won't pull this PR until that AZDO PR is also accepted.

----

@ni/rtos

### DIFF
--- a/scripts/buildVM.sh
+++ b/scripts/buildVM.sh
@@ -29,8 +29,6 @@ Arguments:
   -n name         The name of the final VM archive
   -r recipe_name  The name of the recovery media ISO, within the images_dir
                   Example: \`foo-bar-x64.iso\` has recipe_name: \`foo-bar\`
-  -v              Enable verbose mode. VM serial output will be connected to
-                  stdout.
 EOF
     exit $rc
 }
@@ -43,10 +41,9 @@ initramfsRecipeName=""
 bootDiskSizeMB=""
 memSizeMB=""
 answerFile=""
-verbose_mode=0
 imagesDir=""
 
-while getopts "i:n:r:d:m:a:hv" opt; do
+while getopts "i:n:r:d:m:a:h" opt; do
    case "$opt" in
    a )  answerFile="$OPTARG" ;;
    d )  bootDiskSizeMB="$OPTARG" ;;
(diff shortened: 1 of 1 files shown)


scripts/buildVM.sh
```diff
@@ -55,18 +52,11 @@ while getopts "i:n:r:d:m:a:hv" opt; do
...
    n )  vmName="$OPTARG" ;;
    r )  initramfsRecipeName="$OPTARG" ;;
-   v )  verbose_mode=1 ;;
    \?)  print_usage_and_die ;;
    esac
 done
 shift $(($OPTIND - 1))
 
-if [ "$verbose_mode" -eq 0 ] ; then
-    do_silent() { "$@" &>/dev/null; }
```
shruthi-ravi, line 56: Learning here... Can you explain how redirecting $@ squashes qemu output?

shruthi-ravi (approved):

amstewart:
**NOTICE TO REVIEWERS**
I have run the related changes to the `nilrt_virtual_machines` Makefile target, which accommodate the new `buildVM.sh` API, and verified that it [does not use the `-v` flag](https://dev.azure.com/ni/DevCentral/_build/results?buildId=784922&view=logs&j=44e70749-ceba-5737-fad1-33069546191c&t=a2065b91-4596-5839-9588-2e1a5df2fc02&l=22600) (the fact that the PR run failed elsewhere is expected). I consider this PR to be ready for submission once others have had an opportunity to review.

scripts/buildVM.sh
amstewart: The `do_silent()` function here is just calling the function parameters `$@` as a generic shell command, but with a redirect applied.

```bash
# So...
do_silent qemu-system...
# becomes...
qemu-system ... &>/dev/null
# ... within the context of the function
```

`&>/dev/null` is interpreted by the shell as: *within the context of the previous command/function, open a file pointer to `/dev/null` and copy it into **all open file pointers** `&`*

amstewart: Fun fact that I learned recently: the redirect semantics work the same way for command groups and functions, as they do for single commands. So all of the following are exactly equivalent:
```bash
echo some test text 1>/dev/null

{
  echo some test text
} 1>/dev/null

do_silent_1() {
  $@ 1>/dev/null
}
do_silent_1 echo some test text

do_silent_2() {
  $@
} 1>/dev/null
do_silent_2 echo some test text
```

shruthi-ravi: Good to know. Thanks for the detailed explanation!

jpautler:
I know this is post-merge, but looks good to me also.

PR #46, opened 2021-03-23
azdo: Add script to fetch/push sstate-cache

shruthi-ravi:
This script will be used by the azdo jobs to sync down the sstate
cache from argo at build time, as well as push the local sstate cache
to argo after the build finishes.
The script maintains a maximum of five sstate cache directories on
the remote location and will delete older directories to maintain the
the limit.

PR runs from azdo [here](https://ni.visualstudio.com/DevCentral/_git/ni-central/pullrequest/136867).
Azdo bug reference [here](https://ni.visualstudio.com/DevCentral/_workitems/edit/1299743).

@ni/rtos

### DIFF
--- a/scripts/azdo/update_sstate_cache.sh
+++ b/scripts/azdo/update_sstate_cache.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+
+usage() {
+	local exit_code=${1:-2}
+	test $exit_code -eq 0 || exec 1>&2
+
+	cat <<EOF
+
+Syntax:
+$(basename $BASH_SOURCE) push local_sstate_cache_dir argo_sstate_cache_dir
+   Push the local sstate-cache directory contents to the remote argo
+   server directory.
+
+$(basename $BASH_SOURCE) fetch local_sstate_cache_dir argo_sstate_cache_dir
+   Fetch the latest sstate cache from the remote argo cache directory
+   and install to the local sstate-cache directory.
+
+$(basename $BASH_SOURCE) -h
+   Print this help message and exit
+
+Arguments:
+local_sstate_cache_dir      Path to the local sstate cache directory
+argo_sstate_cache_dir       Path to the remote argo sstate cache directory
+
+EOF
+	exit $exit_code
+}
+
+push_cache() {
+	if [ ! -d $local_sstate_cache_dir -o -z "$(ls -A $local_sstate_cache_dir 2>/dev/null)" ]; then
+		echo $local_sstate_cache_dir is missing/empty. Please provide a valid local_sstate_cache_dir name with sstate-cache contents.
+		usage
+	fi
+
+	mkdir -p $argo_sstate_cache_dir
+	pushd $argo_sstate_cache_dir >/dev/null
+
+	# Copy the contents of the cache to a temp .latest directory
+	# before removing the .latest suffix from the name.
+	temp_argo_dir=$(date +"%s").latest
+	mkdir $temp_argo_dir
+	echo Pushing sstate cache to argo at $temp_argo_dir
+	cp -rL $local_sstate_cache_dir/* $temp_argo_dir/
+	mv $temp_argo_dir ${temp_argo_dir%.latest}
+
+	# Cap the number of cache directories on the argo server at
+	# max_dir_count. Delete older cache directories exceeding the
+	# cap.
+	max_dir_count=5  # arbitrarily chosen number
+	current_dir_count=$(find * -maxdepth 0 -type d 2> /dev/null | wc -l)
+	if [ $current_dir_count -gt $max_dir_count ]; then
+		dirs=$(ls -trd * | head -$(($current_dir_count - $max_dir_count)))
+		for dir in ${dirs[@]}; do
+			echo Removing old sstate cache directory $dir
+			rm -rf $dir
+		done
+	fi
+
+	popd >/dev/null
+}
+
+fetch_cache() {
+	if [ ! -d $argo_sstate_cache_dir ]; then
+		echo Cannot find argo sstate cache directory $argo_sstate_cache_dir
+		usage
+	fi
+
+	pushd $argo_sstate_cache_dir >/dev/null
+
+	mkdir -p $local_sstate_cache_dir
+	# Rsync the latest cache directory, excluding any directory
+	# with a .latest suffix from the search.
+	latest_cache_dir=$(ls -t -I *.latest | head -1)
+	echo Copying over the cache contents from $latest_cache_dir. Might take a while...
+	rsync -rltxSWh --info=STATS1 $latest_cache_dir/ $local_sstate_cache_dir/
+
+	popd >/dev/null
+}
+
+validate_args() {
+	if [ -z $local_sstate_cache_dir ]; then
+		echo Missing local_sstate_cache_dir argument.
+		usage
+	fi
+	if [ -z $argo_sstate_cache_dir ]; then
+		echo Missing argo_sstate_cache_dir argument.
+		usage
+	fi
+}
+
+local_sstate_cache_dir=$2
+argo_sstate_cache_dir=$3
+case "$1" in
+	push)
+		validate_args
+		push_cache
+	;;
+	fetch)
+		validate_args
+		fetch_cache
+	;;
+	--help|-h)
+		usage 0
+	;;
+	*)
+		echo Invalid argument
+		usage
+	;;
+esac
+
--- a/scripts/azdo/update_sstate_cache.sh
+++ b/scripts/azdo/update_sstate_cache.sh
@@ -10,7 +10,8 @@ usage() {
 Syntax:
 $(basename $BASH_SOURCE) push local_sstate_cache_dir argo_sstate_cache_dir
    Push the local sstate-cache directory contents to the remote argo
-   server directory.
+   server directory. Note that this script assumes sequential pushes
+   to the cache.
 
 $(basename $BASH_SOURCE) pull local_sstate_cache_dir argo_sstate_cache_dir
    Fetch the latest sstate cache from the remote argo cache directory
--- a/scripts/azdo/update_sstate_cache.sh
+++ b/scripts/azdo/update_sstate_cache.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
 
 usage() {
 	local exit_code=${1:-2}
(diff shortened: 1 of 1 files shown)


scripts/azdo/update_sstate_cache.sh
```diff
@@ -0,0 +1,110 @@
...
+	mkdir -p $argo_sstate_cache_dir
+	pushd $argo_sstate_cache_dir >/dev/null
+
+	# Copy the contents of the cache to a temp .latest directory
+	# before removing the .latest suffix from the name.
+	temp_argo_dir=$(date +"%s").latest
+	mkdir $temp_argo_dir
+	echo Pushing sstate cache to argo at $temp_argo_dir
+	cp -rL $local_sstate_cache_dir/* $temp_argo_dir/
+	mv $temp_argo_dir ${temp_argo_dir%.latest}
```
shruthi-ravi, line 53: Note that because this is on argo, the `mv` operation may not be atomic, and there is a possibility an incomplete cache could be consumed by build jobs. I could instead:

1. Create a temp file in the local cache directory before starting the `cp` operation, say `donotuse`
2. Remove the file in the remote directory after the `mv` operation has completed 
3. Update the `fetch_cache` operation to ignore any directory with the .latest suffix, as well a `donotuse` file

Lmk what you think.

scripts/azdo/update_sstate_cache.sh
```diff
@@ -0,0 +1,110 @@
...
+Arguments:
+local_sstate_cache_dir      Path to the local sstate cache directory
+argo_sstate_cache_dir       Path to the remote argo sstate cache directory
+
+EOF
+	exit $exit_code
+}
+
+push_cache() {
+	if [ ! -d $local_sstate_cache_dir -o -z "$(ls -A $local_sstate_cache_dir 2>/dev/null)" ]; then
```
amstewart: The `test` here will error in cases where the `$local_sstate_cache_dir` variable contains spaces. You'll want to surround it with double-quotes.

Ditto for most of the `cp` and `mkdir` operations elsewhere in the script.

scripts/azdo/update_sstate_cache.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+
```
amstewart: It's probably best to add `set -eu` here, since I see a few places in the script where operations could fail (`mkdir` on the argo share for example), and we would want to bail out in those cases.

scripts/azdo/update_sstate_cache.sh
```diff
@@ -0,0 +1,110 @@
...
+		usage
+	fi
+
+	mkdir -p $argo_sstate_cache_dir
+	pushd $argo_sstate_cache_dir >/dev/null
+
+	# Copy the contents of the cache to a temp .latest directory
+	# before removing the .latest suffix from the name.
+	temp_argo_dir=$(date +"%s").latest
+	mkdir $temp_argo_dir
```
amstewart, line 49: You should add a trap in the script (`TRAP function EXIT`) to clean up the `.latest` directory, if the script fails during the upload.

scripts/azdo/update_sstate_cache.sh
```diff
@@ -0,0 +1,110 @@
...
+	cat <<EOF
+
+Syntax:
+$(basename $BASH_SOURCE) push local_sstate_cache_dir argo_sstate_cache_dir
+   Push the local sstate-cache directory contents to the remote argo
+   server directory.
+
+$(basename $BASH_SOURCE) fetch local_sstate_cache_dir argo_sstate_cache_dir
+   Fetch the latest sstate cache from the remote argo cache directory
+   and install to the local sstate-cache directory.
```
amstewart: MINOR: I think the commands should be `push` and `pull` (rather than `fetch`), in keeping with `git` terminology.

scripts/azdo/update_sstate_cache.sh
amstewart: >Update the fetch_cache operation to ignore any directory with the .latest suffix, as well a donotuse file

What about using a suffix like `*.tmp` on the argo directory while you're uploading the content and then removing that `.tmp` suffix when it is ready for consumption. I think the pull/fetch operation should avoid any directory ending in `.tmp` and simply find the latest, non-temp directory using their timestamps.

The local directory operations you talk about (`donotuse`) are unnecessary in that case.

scripts/azdo/update_sstate_cache.sh
```diff
@@ -0,0 +1,110 @@
...
+	# before removing the .latest suffix from the name.
+	temp_argo_dir=$(date +"%s").latest
+	mkdir $temp_argo_dir
+	echo Pushing sstate cache to argo at $temp_argo_dir
+	cp -rL $local_sstate_cache_dir/* $temp_argo_dir/
+	mv $temp_argo_dir ${temp_argo_dir%.latest}
+
+	# Cap the number of cache directories on the argo server at
+	# max_dir_count. Delete older cache directories exceeding the
+	# cap.
```
amstewart, line 59: As this block is written, it just deletes any content in argo. It should discriminate between the temp/latest directories and the _complete_ directories.

Also, I think you're relying on `find` to sort the directories, and I don't think that's reliable. I'm not sure it will always delete the oldest timestamp.

scripts/azdo/update_sstate_cache.sh
shruthi-ravi: I think the current script does what you described, albeit using a .latest suffix and not .tmp? Is there a different mechanism to remove the suffix if not with the `mv` command?
With `mv`, I was thinking that if the sequence turned out such that a new directory (with the suffix removed) gets created before the contents of the old directory are moved to the new directory, there is a chance that when fetch is called it picks up the new directory because it doesn't have the suffix, but the cache contents in the new directory would not be complete for a while.

scripts/azdo/update_sstate_cache.sh
shruthi-ravi: There shouldn't be any .latest directories at this point. A .latest directory if created before would have already been renamed above.
I'm relying on `find` to count the number of directories before proceeding to delete. I could have used `ls`, but I didn't think of it then and don't think it will change the result now. I use `ls` later on to sort (by file modification time, not the name of the directory) and pick the directories to remove.

amstewart: >There shouldn't be any .latest directories at this point.

What if two pushes are being run concurrently?

>I use ls later on to sort (by file modification time, not the name of the directory) and pick the directories to remove.

Hmm.. I'm not sure we can rely on file modification time across the CIFS boundary. Since they're already named using UNIX-time, why not rely on that?

shruthi-ravi: > What if two pushes are being run concurrently?

Only the CI jobs will attempt pushes, and they are scheduled to run sequentially.

>Hmm.. I'm not sure we can rely on file modification time across the CIFS boundary. Since they're already named using UNIX-time, why not rely on that?

Sure.

scripts/azdo/update_sstate_cache.sh
shruthi-ravi: @amstewart what do you think?

amstewart: I think I got a little lost in the discussion last week. I have a more-clear picture of what you're talking about now and I think the `donotuse` poison-pill idea is a pretty elegant solution. ACK from me. :+1:

scripts/azdo/update_sstate_cache.sh
amstewart: >Only the CI jobs will attempt pushes, and they are scheduled to run sequentially.

I agree that *shouldn't* happen, but I'm uncomfortable relying on that assumption. I'll defer to you as to whether or not the enhancement is worth it.

shruthi-ravi:
**Updates**
v2: 
- Add `set -e`
- Replace `fetch` with `pull`
- Use quotes when referring to the argo cache and local sstate cache
  dirs
- Add a trap to delete the .latest directory if things go wrong
- Create a DONOTUSE file in the remote cache directory location
  before copying the cache in a push operation. This is to safeguard
  against pulling the cache directory before the mv operation can be
  completed, which could result in an incomplete local cache.
  The pull operation will now only fetch the latest directory from
  argo that does not have the .latest suffix as well as the DONOTUSE
  file.
- Sort and remove older cache directories on argo based on the name
  of the directories instead of file modification time.

v3:
- Add note about the sequential push operations assumption this script
  makes. There are several changes required to support parallel push 
  operations. To keep it simple and support the current use case where
  CI builds are run sequentially, stick to that assumption.

scripts/azdo/update_sstate_cache.sh
shruthi-ravi: Note that I didn't add the `-u` option because I already run checks for arguments to print helpful error messages and the script starts to error out much earlier when the `-u` option is set.

scripts/azdo/update_sstate_cache.sh
```diff
@@ -0,0 +1,125 @@
...
+
+	# Copy the contents of the cache to a temp .latest directory
+	# before removing the .latest suffix from the name. Create a
+	# DONOTUSE file in the .latest directory until the mv operation is
+	# complete. The pull operation will not rsync a folder with this
+	# file. Implement this extra precaution because the mv may not be
+	# atomic on the argo share.
+
+	temp_argo_dir=$(date +"%s").latest
+	trap "{ rm -rf $temp_argo_dir; }" EXIT
```
amstewart, line 48: `TRAP`s have scope until the process exits, so this `TRAP` will be in place for the entirety of this script. It's probably inconsequential - since `$temp_argo_dir` should not exist for very long anyway - but it would be cleaner if you unset the TRAP after the target directory gets renamed (line 49). You can use the provisioning test script I wrote recently [as a reference](https://github.com/ni/nilrt/blob/nilrt/master/dunfell/scripts/tests/provisioning/start-vm-using-expect.sh#L126) for how to do that.

amstewart (changes requested):
lgtm; apart from the TRAP sticking around for longer than it should.

amstewart (approved):
ACK. Looks good to me. :+1:

shruthi-ravi:
Ugh, gross! I wanted to squash my commits before rebasing. Can't fix it now can I? :( @amstewart

shruthi-ravi:
I'm probably going to stick to force pushes..

amstewart:
I don't think it's worth fixing. The commit requirements for `nilrt.git` are relatively loose, since it has no other upstream and doesn't get rebased.

PR #47, opened 2021-03-23
test_migration_to_rauc: Trust custom feeds

chaitu236:
@ni/rtos


shruthi-ravi (approved):

jpautler (approved):
My only suggestion might be to provide a little more detail in the commit message.

chaitu236:
> My only suggestion might be to provide a little more detail in the commit message.

Done

amstewart (approved):

PR #48, opened 2021-03-24
expect files: Increase timeout

chaitu236:
@ni/rtos 

NOTE: Testing on azdo still in progress, but it appears that azdo failures were due to timeout values being too low for non-kvm VMs. It is not clear if console closing on reboot is also part of the problem but increasing the timeout without making any other changes fixes the issue when run locally.

### DIFF
--- a/scripts/tests/provisioning/migrate_vm_to_safemode.expect
+++ b/scripts/tests/provisioning/migrate_vm_to_safemode.expect
@@ -33,7 +33,7 @@ expect {
 	}
 }
 
-set timeout 90
+set timeout 180
 # configure opkg
 expect {
 	-re "$USER@.*# $" {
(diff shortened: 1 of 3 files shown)


gratian (approved):
The commit message seems truncated/missing signoff, otherwise LGTM.

chaitu236:
> The commit message seems truncated/missing signoff, otherwise LGTM.

Fixed. Thanks.

shruthi-ravi (approved):

jpautler (approved):
It still doesn't make any sense to me why the timeout was being triggered before the old timeout period had elapsed, but I guess I can't argue if it's working now. :)

amstewart (approved):

amstewart:
@chaitu236 Let me know when you're done verifying your changes.

PR #51, opened 2021-04-02
buildUSB.sh: Add script to create bootable usb

billpittman:
Add script to build a bootable USB stick that can be
used to boot a target instead of using the internal
non-volatile storage.

Signed-off-by: Bill Pittman <bill.pittman@ni.com>

### DIFF
--- a/scripts/buildUSB.sh
+++ b/scripts/buildUSB.sh
@@ -1,0 +1,163 @@
+#!/bin/bash
+set -euo pipefail
+
+IMAGE_FILE="../build/tmp-glibc/deploy/images/x64/nilrt-base-bundle-image-x64.tar.bz2"
+DISK_CONFIG="../sources/meta-nilrt/recipes-core/initrdscripts/files/disk_config_x64"
+ASK_FIRST=true
+
+if [ -e ${DISK_CONFIG} ] ; then
+   source ${DISK_CONFIG}
+else
+   error_and_die "disk_config_x64 not found"
+fi
+
+error_and_die () {
+    echo >&2 "ERROR: $1"
+    exit 1
+}
+
+cleanup () {
+    local exitCode="$?"
+
+    set +e
+
+    if [ -e "${A_MOUNT:-}" ] ; then
+        umount ${A_MOUNT} >/dev/null 2>&1
+        rm -fr ${A_MOUNT}
+    fi
+
+    exit "$exitCode"
+}
+
+trap cleanup EXIT
+
+print_usage () {
+cat <<EOF
+Usage: buildUSB.sh -h|-l|[-i boot-image-file] [-y] USB_DEVICE
+
+Options:
+  -h         Print this help message and exit
+  -l         List valid removable devices and exit
+  -i         specify path to boot image file
+  -y         do not ask for confirmation
+  USB_DEVICE use specified block device (ie /dev/sdX)
+EOF
+}
+
+print_usage_and_die () {
+    local rc=${1:-2}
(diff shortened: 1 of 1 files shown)


billpittman:
Just in case you want to try to script, but don't want to read the QEMU docs, here's how to pass a USB stick from the host to a qemu VM that we export.

sudo ./run-nilrt-vm-x64.sh -c 2 -s -b virbr0 -- -device qemu-xhci,id=xhci -device usb-host,vendorid=<VID>,productid=<PID>
<VID> - in the form 0xdead
<PID> - in the form 0xbeef

If you press ESC during boot (just before GRUB runs), you'll be presented with a boot menu and can choose the USB to boot from -- Thanks Jeff for pointing this out.

billpittman:
Attn: @ni/rtos

amstewart (changes requested):
Neat. And I'm fine with this going into the dunfell sources, but I'd like to ensure that someone who is browsing the code won't confuse this for a supported live-USB workflow.

Could you add an `echo` to the top of the file with something like.
```
WARNING: $(basename $0) is not an NI-supported script and is only appropriate for experimental use. Use at your own risk.
```

amstewart:
>- in the form 0xdead
>- in the form 0xbeef

I've seen that PID/VID before. Apart from being hillarious, does it have any significance to us? Do we switch code based on it?

billpittman:
> > * in the form 0xdead
> > * in the form 0xbeef
> 
> I've seen that PID/VID before. Apart from being hillarious, does it have any significance to us? Do we switch code based on it?

No significance at all, I was just trying to show format.  The actual VID/PID need to match the USB stick in question, but the qemu command line was pretty picky about having to specify hex notation in the command line - just trying to make things easier for whoever wanted to try it.

billpittman:
> Neat. And I'm fine with this going into the dunfell sources, but I'd like to ensure that someone who is browsing the code won't confuse this for a supported live-USB workflow.
> 
> Could you add an `echo` to the top of the file with something like.
> 
> ```
> WARNING: $(basename $0) is not an NI-supported script and is only appropriate for experimental use. Use at your own risk.
> ```

I don't have a problem changing comments/output, but honestly I expect manufacturing to use this feature (since they requested it), and I don't feel a "Use at your own risk" statement should be part of any manufacturing process.

amstewart:
> I don't have a problem changing comments/output, but honestly I expect manufacturing to use this feature (since they requested it), and I don't feel a "Use at your own risk" statement should be part of any manufacturing process.

Ah. Maybe "for internal use only", then?

billpittman:
Added a warning as requested.

scripts/buildUSB.sh
```diff
@@ -0,0 +1,163 @@
...
+verify_continue $target || error_and_die "Aborted!"
+
+echo "Partitioning and formating $target"
+disk_setup $target >/dev/null 2>&1 || error_and_die "You may need root permission for formatting/partitioning the usb device."
+
+echo "Copying files to $target (this will take a few minutes)"
+
+# Create temporary mount point
+readonly A_MOUNT=$(mktemp -d "/tmp/niboota-XXXXXXX")
+mount "${target}1" ${A_MOUNT}
```
shruthi-ravi, line 150: Why the 1 in "${target}**1**"?

billpittman: Because the script needs to mount the first partition (that the script just created), so /dev/sdc1 for example.  Target actually holds the value of the device block device itself (not a partition) so /dev/sdc for example.

amstewart (approved):

PR #53, opened 2021-04-07
Deprecate meta-orgconf in sumo

amstewart:
To deprecate meta-orgconf, some of the org-specific variables it defines need to be moved to an optional conf snippet in nilrt.git. This PR essentially merges the changes to the init env file we have already upstreamed to the dunfell mainline, and adds a similar `ni-org.conf` snippet that we have upstreamed to the same.

Builders can use the org snippet using the same `-o/--org` argument to the init script.

## Testing
* Verified that the new init script works, both with and without the `-o` arg.
* Verified that bitbake can parse the configuration files and that the variables look correct.

----
I expect this PR and the associated meta-nilrt PR will necessitate changes to NIOE to get them building, I'll work on those next.

Ref meta-nilrt [PR 159](https://github.com/ni/meta-nilrt/pull/159).

@ni/rtos

### DIFF
--- a/ni-oe-init-build-env
+++ b/ni-oe-init-build-env
@@ -1,8 +1,51 @@
-TEMPLATECONF=${TEMPLATECONF:-$PWD/sources/meta-nilrt/conf}
-BITBAKEDIR=${BITBAKEDIR:-$PWD/sources/bitbake}
-BUILDDIR=${BUILDDIR:-$PWD/build}
-BB_ENV_EXTRAWHITE=${BB_ENV_EXTRAWHITE:-ENABLE_BUILD_TAG_PUSH}
+#!/usr/bin/env
 
-export PS1="(bb) $PS1"
+SCRIPT_ROOT=$(realpath $(dirname ${BASH_SOURCE:-${0}}))
+
+enable_ni_org_conf=false
+positionals=()
+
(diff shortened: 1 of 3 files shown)


shruthi-ravi (approved):

ni-oe-init-build-env
```diff
@@ -1,8 +1,51 @@
...
+# into the build workspace, if one needs to be created.
+TEMPLATECONF=${TEMPLATECONF:-${SCRIPT_ROOT}/sources/meta-nilrt/conf}
 
-export PS1="(bb) $PS1"
 export TEMPLATECONF
-source $PWD/sources/openembedded-core/oe-init-build-env
+
+# Call OE-upstream's build env initialization script, which will create a build
+# workspace called either "${1:-build}/" and `cd` into it.
+cd ${SCRIPT_ROOT}
```
billpittman, line 33: Not 100% sure why the cd is necessary ... is it to cover the possibility of someone calling ni-oe-init-build-env without first being in the same directory as the script?

amstewart: It's been a while since I wrote that line, but iirc. it is to coerce the `build/` directory to be created within the top level of the nilrt repo. Our bitbake configurations don't behave predictably when the build workspace is in another location.

billpittman (approved):

PR #59, opened 2021-04-20
More migration test updates

amstewart:
This patchset contains additional fixups and modifications which are necessary as part of the effort to enable the `rtos_nifeeds` build component and close out the backwards migration tests.

Primary changes:
* Reconcile the RAUC provisioning tcl expect script with the new expect script arguments.
* Rename the provisioning scripts so that they don't talk about migration.(I actually didn't notice that the previous implementation didn't work because I got confused by the old names.)
* Change the login user from `root` to `admin`, now that NIAuth is available.

These changes are needed before [AZDO PR 148906]() can be pulled.

## Testing
* The forward and backward migration test targets are both building on my dev machine, using these new scripts.

@ni/rtos


amstewart:
## Update 2
* Added a fix to the migrate-to-safemode script. When I switched the `$USER` value from `root` to `admin`, it broke detection of the recovery USB prompt - which is always `root@recovery...`.

scripts/tests/provisioning/test_rauc_provision.expect
```diff
@@ -10,13 +10,15 @@
...
+set ipk_feeds [lrange $argv 2 end]
 # The SLIRP networking static ipv4 to use for the VM host (as viewed from the
 # guest). Can be any ip on the 10.0.2.0/24 subnet. 10.0.2.2 is the QEMU
 # default, if none is specified.
 set vm_host_ip "10.0.2.2"
 # Set the login username
-set USER root
+set USER admin
+
+set RE_PROMPT "(admin|root)@.*# $"
```
chaitu236, line 21: Why do we need to check for either? We're already setting USER to admin and using admin for logging so if we receive root prompt instead of admin, it'd be wrong

scripts/tests/provisioning/migrate_vm_to_safemode.expect
```diff
@@ -83,7 +83,7 @@ expect {
 
 set timeout 500
 expect {
-	"$USER@recovery:/#" {}
+	"root@recovery:/#" {}
```
chaitu236, line 86: Can we just re-set USER to root and use it?

chaitu236 (changes requested):

scripts/tests/provisioning/test_rauc_provision.expect
amstewart: I'm just checking for both because they are equivalent on NILRT and I think it's a pain to have to go through and swap them around every time we change the login $USER.

scripts/tests/provisioning/migrate_vm_to_safemode.expect
amstewart: Well, I don't think there's a reason to variablize it, since the recovery USB will *never* use `admin` as its login (because it necessarily doesn't include NIAuth).

chaitu236 (approved):

PR #60, opened 2021-04-22
tests/provisioning: restrict feed access and empower test wrapper scripts

amstewart:
This patchset does two primary things:
1. It removes the `base-feed.conf` opkg sources prior to installing the newly build grub-gateway and provisioning test scripts. This keeps the tests fixtures from accidently downloading old (already published) feed content, if the locally-built feeds don't have what is necessary. We would prefer for the test to just fail in that case, rather than falsely passing.
2. It adds a wrapper script for running the migration tests, and empowers it - and the extant `test_provisioning.sh` script - to pass the lower-leverl tcl-expect script the files they need, and to parse stdout for the test results. The goal here is to (a) allow external builders to easily call these same wrapper scripts and (b) to generally move test logic out of the AZDO component. There will be an associated AZDO PR, which interfaces with the new wrapper scripts.

## Testing
Both test_ targets build on my machine.

@ni/rtos

PR #61, opened 2021-04-23
[Patch V2] tests/provisioning: restrict feed access and empower test wrapper scripts

amstewart:
It would seem that Github didn't like my [previous PR](https://github.com/ni/nilrt/pull/60), and for some reason would not accept new commits to the merge branch. This PR is a replacement, which includes the version 2 commits.

---

This patchset does two primary things:

1. It removes the base-feed.conf opkg sources prior to installing the newly build grub-gateway and provisioning test scripts. This keeps the tests fixtures from accidently downloading old (already published) feed content, if the locally-built feeds don't have what is necessary. We would prefer for the test to just fail in that case, rather than falsely passing.
2. It adds a wrapper script for running the migration tests, and empowers it - and the extant test_provisioning.sh script - to pass the lower-leverl tcl-expect script the files they need, and to parse stdout for the test results. The goal here is to (a) allow external builders to easily call these same wrapper scripts and (b) to generally move test logic out of the AZDO component. There will be an associated AZDO PR, which interfaces with the new wrapper scripts.

Testing

Both test_ targets build on my machine.

---

## Version 2
* Also remove the /etc/opkg/NI-dist.conf feed source for the same reason. [commit](https://github.com/ni/nilrt/pull/61/commits/271f22e0e747311deacc1bfb87a5e56f6087af68).

Testing
* Rebuilt test_ targets on my dev machine and confirmed that they still work and look good.

@ni/rtos

### DIFF
--- a/scripts/tests/provisioning/test_grub_gateway_migration.sh
+++ b/scripts/tests/provisioning/test_grub_gateway_migration.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+#
+# This script should be run to verify the functionality of the
+# dist-nilrt-grub-gateway IPK.
+# This test:
+#  * boots an already-provisioned NILRT QEMU virtual machine, using the RAUC
+#    (efi-ab) bootflow.
+#  * configures opkg with a locally-hosted ipk feed
+#  * installs and run the dist-nilrt-grub-gateway IPK, to reprovision the VM as
+#    a legacy safemode (grub) image
+#  * reconfigures opkg with the locally-build dist/ feed
+#  * installs and runs the grub-gateway ptest suite
+#
+#  This script returns 0 when the grub-gateway ptest suite passes, or a
+#  positive value otherwise.
+set -euo pipefail
+#set -x
+
+SCRIPT_ROOT=$(dirname "$BASH_SOURCE")
+
+usage () {
+	cat <<EOF
+$(basename $0) nilrt_vm_directory dist_feed_path other_ipk_feeds
+	       [other_ipk_feeds...]
+EOF
+	exit ${1:-2}
+}
+
+if [ $# -lt 3 ]; then
+	echo "ERROR: invalid or missing arguments." >&2
+	usage
+fi
+
+vm_dir=${1}
+dist_feed=${2}
+ipk_feeds=(${@:3})
+log_file=./$(basename ${BASH_SOURCE%.*}).log
+
+(
+	bash "${SCRIPT_ROOT}/start-vm-using-expect.sh" \
+		${ipk_feeds[@]/#/--ipk-feed /} \
+		--ipk-feed "${dist_feed}" \
+		"${vm_dir}" \
+		migrate_vm_to_safemode.expect
+
+	bash "${SCRIPT_ROOT}/start-vm-using-expect.sh" \
+		--ipk-feed "${dist_feed}" \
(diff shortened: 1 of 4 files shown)


billpittman (approved):

scripts/tests/provisioning/test_provisioning.sh
```diff
@@ -9,7 +9,29 @@
 #  * evaluates the ptest results.
 
 script_dir="$(dirname "$0")"
-source "$script_dir/start-vm-using-expect.sh"
```
shruthi-ravi, line 12: Do you need to add `set -euo pipefail` here?

scripts/tests/provisioning/test_grub_gateway_migration.sh
```diff
@@ -0,0 +1,63 @@
...
+		${ipk_feeds[@]/#/--ipk-feed /} \
+		--ipk-feed "${dist_feed}" \
+		"${vm_dir}" \
+		migrate_vm_to_safemode.expect
+
+	bash "${SCRIPT_ROOT}/start-vm-using-expect.sh" \
+		--ipk-feed "${dist_feed}" \
+		"${vm_dir}" \
+		test_safemode_provision.expect
+) | tee $log_file
```
shruthi-ravi, line 49: So are there going to be two log files for each test now? start-vm-using-expect.sh also [passes](https://github.com/ni/nilrt/blob/nilrt/master/dunfell/scripts/tests/provisioning/start-vm-using-expect.sh#L145) in a log file name to the expect files. I expected an accompanying change to stop doing that, but maybe I've misunderstood the big picture.

scripts/tests/provisioning/test_provisioning.sh
amstewart: That's a good idea. Will fix.

scripts/tests/provisioning/test_grub_gateway_migration.sh
amstewart: That's correct. These new wrapper scripts aren't using that functionality anymore. I didn't bother removing it from the `start-` script because it isn't hurting anything and could potentially be useful to someone using that script directly.

amstewart:
## Version 3
* Add shell options to the `test_provisioning.sh` script, for safety.

scripts/tests/provisioning/test_grub_gateway_migration.sh
shruthi-ravi: Can you share what the log file names all look like if you have a run on your system?

amstewart: For the migration test:
```
migrate_vm_to_safemode.log  # created by start- script; contains the VM session stdout from the tcl expect script of the same name
test_safemode_provision.log  # created by start- script; contains the VM session stdout from the tcl expect script of the same name
test_grub_gateway_migration.log  # created by the wrapper script; contains a cat of the above two, plus the setup/teardown done in the start- script
```

The test_provisioning wrapper creates similar - but there is only one VM session log.

shruthi-ravi: Thanks. IMO the multiple logs and their names are confusing. But we can fix it later if someone complains enough!

shruthi-ravi (approved):

PR #72, opened 2021-08-17
Revert "gitmodules: remove meta-measured"

amstewart:
This reverts commit 26ace02a6eeb00aaee3670c79e13f6318e89e88b.

Add meta-measured back to the layer stack, and set its upstream head to
`nilrt/master/hardknott`. We cannot efficiently drop meta-measured until
we put more work into meta-nilrt, to drop its dependencies.


amstewart:
@ni/rtos

shruthi-ravi (approved):

PR #73, opened 2021-08-20
scripts/docker: upgrade Dockerfile for hardknott

amstewart:
The Dockerfile in scripts/docker provides a limited container
implementation which contains the host dependencies needed to build
NILRT.

Replace it with a more complete solution, based on Debian 10.10, which
is appropriate for building NILRT on the hardknott mainline. Also
include a docker-compose file and shell script entrypoint, which
provides "one command" entry into a viable build workspace.

---

To run the container:
```bash
# from the nilrt.git repo...
bash ./scripts/docker/run.build-nilrt.sh
# from within the container...
source ni-oe-init-build-env
# you're in bitbake
```

@ni/rtos


chaitu236 (approved):

scripts/docker/run.build-nilrt.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
shruthi-ravi, line 1: Why the first '.' in `run.build-nilrt.sh`? Is that some specific file name format you are following?

shruthi-ravi (approved):

scripts/docker/run.build-nilrt.sh
amstewart: Nope; that's just how I like to name stuff.

PR #75, opened 2021-08-25
Rebase dunfell mainline commits to hardknott

amstewart:
To capture the progress we made on the dunfell mainline in hardknott, rebase the dunfell mainline onto hardknott.

The major efforts in the dunfell mainline were:
1. Adding the CHANGELOG - Dropped all these changes, because they were already reconciled with sumo.
2. Adding AZDO "dunfell" pipeline configuration snippets and updating the `init-` sourcefile - dropped; already reconciled with sumo.
2. Adding the fio tests - picked.
3. Documentation cleanup (README.md, CONTRIBUTING, "building the kernel" tutorial, et c.) - picked.
4. Reconfiguration the QEMU VMs for RAUC - retained, with the understanding that we will undo some of these changes as we're getting the new pipeline online. (I can't easily test changes to the provisioning tool until we have the pipeline work done.)
5. Adding the RAUC provisioning tests - retained, because they are harmless to retain in the source and might provide a good basis for the equivalent hardknott tests.

## Testing
These are mostly non-functional changes, but I confirmed that I can still enter a bitbake build environment. I'm sure the `buildVM` tool is broken, but we need to touch that workflow as part of building out the pipeline anyway.

@ni/rtos


shruthi-ravi (approved):

amstewart:
## Patch v2
* reverted the `buildVM` `efi-ab` boot scheme commit - since it is the most likely to conflict with safemode/runmode provisioning.

PR #76, opened 2021-08-30
scripts: add pr-server docker container

amstewart:
scripts: add pr-server docker container
    
    Add a docker container and compose spec which can be used to quickly
    standup a bitbake PatchRev server using an arbitary bitbake repo
    upstream ref.

----

This container is going to be useful when we eventually *kubernate* the natinst PR-server. But this is independently useful for external builders and developers, if they find themselves with a need to setup a particular version of the PR server.

## Testing
* When running on my dev machine, I can setup a sibling `nilrt-build` container and configure it with `PRSERV_HOST = "localhost:8585"`. Everything works as you would expect.

@ni/rtos

### DIFF
--- a/scripts/azdo/pr-server/README.md
+++ b/scripts/azdo/pr-server/README.md
@@ -1,0 +1,38 @@
+# NILRT PatchRev Server Container
+
+## Starting the Server
+
+The PR server can be started with sensible defaults through the normal `docker-compose up` workflow.
+
+```bash
+docker-compose -f pr-server.compose.yml up pr-server
+```
+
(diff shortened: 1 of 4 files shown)


shruthi-ravi:
Just curious, when did you find the need to use this setup?

amstewart:
I created it while trying to figure out if there was a breaking API change between bitbake_1.30 and bitbake_1.50. I wanted a quick way to startup PRserver instances using arbitrary refs.

shruthi-ravi (approved):

PR #77, opened 2021-09-10
scripts: add cve-to-csv.py script

amstewart:
The cve_check class in OE will produce a directory of information about
what reported CVEs are outstanding in the layer stack. The CVE files are
presented in plain-text stanzas, which are not very easy to sort and
filter.

cve-to-csv.py will automatically digest the cve_check directory into a
comma separated value data structure - which it writes to stdout.

### DIFF
--- a/scripts/dev/cve-to-csv.py
+++ b/scripts/dev/cve-to-csv.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+from argparse import ArgumentParser
+from pathlib import Path
+import os
+import sys
+
+
+parser = ArgumentParser(
+    description='Parses a directory full of OE cve_check stanza files, and writes thier csv equivalent to stdout.',
+    )
+parser.add_argument('cve_directory', type=Path,
+    help='Path to the cve_check output directory.'
+    )
+args = parser.parse_args()
+
+
+class CVE():
+
+    FIELDS = {
+        'LAYER': 'layer',
+        'PACKAGE NAME': 'package_name',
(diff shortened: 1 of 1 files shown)


shruthi-ravi (approved):
We publish CVE reports during a build, right? What format are they in?
Do you intend to extend this script to run as part of a post-build step?

amstewart:
@shruthi-ravi 
>We publish CVE reports during a build, right?

We do not at the moment. But we would at least build them if we turned on the `cve_check` INHERIT.

>What format are they in?

They're stored in `:.../tmp-glibc/deploy/cve/*` as plain text files with stanzas like:
```
LAYER: meta
PACKAGE NAME: acl
PACKAGE VERSION: 2.2.53
CVE: CVE-2009-4411
CVE STATUS: Patched
CVE SUMMARY: The (1) setfacl and (2) getfacl commands in XFS acl 2.2.47, when running in recursive (-R) mode, follow symbolic links even when the --physical (aka -P) or -L option is specified, which might allow local users to modify the ACL for arbitrary files or directories via a symlink attack.
CVSS v2 BASE SCORE: 3.7
CVSS v3 BASE SCORE: 0.0
VECTOR: LOCAL
MORE INFORMATION: https://nvd.nist.gov/vuln/detail/CVE-2009-4411
```

>Do you intend to extend this script to run as part of a post-build step?

Not at the moment. But if we turn on `cve_check` and start storing them as artifacts, we might prefer to store the .csv version instead.

PR #78, opened 2021-09-15
Dockerfile: General updates, additions, removals.

jeminor:
Update container to something a small amount newer, drop a few
things we don't use in the build directly, and add some thing we
might need for building a kernel.

Signed-off-by: James Minor <james.minor@ni.com>

### DIFF
--- a/scripts/docker/Dockerfile
+++ b/scripts/docker/Dockerfile
@@ -1,14 +1,64 @@
-FROM debian:9.8
+## LAYER base OS
+FROM debian:9.11 as os-base
 
 RUN rm /bin/sh && ln -s /bin/bash /bin/sh
 
+# Enable contributor package sources
+ARG DEBIAN_FRONTEND=noninteractive
 RUN sed -i 's/stretch main/stretch main contrib/' /etc/apt/sources.list
-
-ENV DEBIAN_FRONTEND=noninteractive
-
 RUN apt-get update && apt-get install --assume-yes apt-utils
 
+# General build niceties
+RUN apt-get update && apt-get install --assume-yes \
+	binutils \
+	bvi \
+	bzip2 \
+	coreutils \
+	git \
+	python3 \
+	rsync \
+	strace \
+	sudo \
+	tree \
+	unzip \
+	vim \
+	wget \
+	xz-utils \
+	zip \
+""
+
+RUN apt-get update && apt-get install --assume-yes \
+	build-essential \
+	ccache \
+	chrpath \
+	cvs \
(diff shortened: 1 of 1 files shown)


scripts/docker/Dockerfile
```diff
@@ -1,21 +1,79 @@
...
+	zip \
+""
+
+# Install nibuild dependencies
+RUN dpkg --add-architecture i386
+RUN apt-get update && apt-get install --assume-yes \
+	libc6-i686:i386 \
+	libperl5.24:i386 \
+	libstdc++6:i386 \
+""
```
shruthi-ravi: Is this needed in this version of the Dockerfile? Given that you have the Azdo version, that seems like the right place for this.

scripts/docker/Dockerfile
```diff
@@ -1,21 +1,79 @@
...
+RUN update-locale LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8
+ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
 
-RUN apt-get update && apt-get install --assume-yes apt-utils
+
+## LAYER build env
+FROM os-base as nilrt-build
 
 # Install OE dependencies
+ARG DEBIAN_FRONTEND=noninteractive
```
shruthi-ravi, line 61: Do you need this line repeated here?

scripts/docker/Dockerfile
```diff
@@ -28,55 +86,51 @@ RUN apt-get update && apt-get install --assume-yes \
...
-	rsync \
 	socat \
 	strace \
 	subversion \
-	sudo \
 	texi2html \
 	texinfo \
 	unzip \
 	uuid-runtime \
-	vim \
```
shruthi-ravi, line 54: Why drop vim? :(

shruthi-ravi: I'd prefer to add to this tool list, and not drop rsync, sudo, etc.

scripts/docker/Dockerfile
jeminor: It failed without that, I suspect since I'm using ARG insted of ENV.  A few articles suggested ARG since that way any user of the container won't be stuck in noninteractive mode if they try to install anything, but I'm guessing it only applied for the current stage (and thus needs repeating at the next stage).

@amstewart let me know if I'm wrong here...

amstewart (approved):

shruthi-ravi (approved):

PR #79, opened 2021-09-16
ni-oe-init-build-env: autocalculate BB_NUMBER_THREADS

amstewart:
The BB_NUMBER_THREADS variable determines the maximum number of bitbake
tasks which are allowed to execute contemporaneously.

If the user asserts a value for this variable, pass it through to
bitbake. Otherwise, calculate the value by doubling the core count -
which is the recommended rule of thumb from the yocto manual.

----

`nproc` is part of the `coreutils` package on most distros - so I consider this to be pretty safe. If `nproc` is not available for some reason, the command (and script) will fail and say as much.

The default value for this variable in OE is set to the core count (x1), by bitbake.conf.
```
#   set? /mnt/workspace/sources/openembedded-core/meta/conf/bitbake.conf:802
#     "${@oe.utils.cpu_count()}"
```

When you execute the script, output looks like this:
```bash
INFO: BB_NUMBER_THREADS not set. Calculating...
BB_NUMBER_THREADS=8
```

## Testing
Tested on my local hardknott docker containers.

@ni/rtos

### DIFF
--- a/ni-oe-init-build-env
+++ b/ni-oe-init-build-env
@@ -21,9 +21,21 @@ BITBAKEDIR=${BITBAKEDIR:-${SCRIPT_ROOT}/sources/bitbake}
 BUILDDIR=${BUILDDIR:-${SCRIPT_ROOT}/build}
 
 BB_ENV_EXTRAWHITE="${BB_ENV_EXTRAWHITE} \
+	BB_NUMBER_THREADS \
 	ENABLE_BUILD_TAG_PUSH \
 	GIT_REPODIR \
 "
+
+# Users can assert the maximum number of concurrent bitbake threads via the
+# BB_NUMBER_THREADS variable, or let this script calculate a reasonable value
+# based on the number of cores.
+if [ -z "$BB_NUMBER_THREADS" ]; then
+	echo "INFO: BB_NUMBER_THREADS not set. Calculating..."
+	# num_cores * 2 is the recommended value from the yocto manual
+	BB_NUMBER_THREADS=$(($(nproc) * 2))
+fi
+export BB_NUMBER_THREADS
+echo "BB_NUMBER_THREADS=${BB_NUMBER_THREADS}"
 
 # Define GIT_REPODIR as the directory containing the OE layer submodule repos.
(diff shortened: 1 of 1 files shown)


ni-oe-init-build-env
```diff
@@ -21,10 +21,22 @@ BITBAKEDIR=${BITBAKEDIR:-${SCRIPT_ROOT}/sources/bitbake}
...
 	GIT_REPODIR \
 "
 
+# Users can assert the maximum number of concurrent bitbake threads via the
+# BB_NUMBER_THREADS variable, or let this script calculate a reasonable value
+# based on the number of cores.
+if [ -z "$BB_NUMBER_THREADS" ]; then
+	echo "INFO: BB_NUMBER_THREADS not set. Calculating..."
+	# num_cores * 2 is the recommended value from the yocto manual
+	BB_NUMBER_THREADS=$(($(nproc) * 2))
```
shruthi-ravi, line 35: If `nproc` is not found, wouldn't it be better to print, and continue with OE's default value?

amstewart: I really can't imagine there will be a scenario where a user *intentionally* doesn't have nproc in the PATH. I'd expect that the user would rather that the script fail and indicate that something is wrong. Can you think of a reasonable host setup that doesn't have nproc?

bstreiff: `nproc` is part of GNU coreutils, which means you'd probably have to be on a distro with a busybox-based userland to not have it, or on a non-Linux system (macOS does not have it, I dunno about the BSDs), and I think if you're in that situation you're in a place that upstream OE doesn't support given the [build host requirements](https://www.yoctoproject.org/docs/latest/mega-manual/mega-manual.html#brief-compatible-distro).

shruthi-ravi: > Can you think of a reasonable host setup that doesn't have nproc?

No idea :). I'm just saying the error handling can be better. Instead of a nproc not found error, maybe just tell the user to set the BB_NUMBER_THREADS var. But maybe it's all a NOOP given Brandon's statement above.

shruthi-ravi (approved):

PR #84, opened 2021-10-14
Use Pyrex as a build tool

amstewart:
This PR...
1. adds a new submodule to the project `sources/pyrex` - which is based on the [garmin/pyrex](https://github.com/garmin/pyrex/) project. Pyrex is a tool which creates a shim layer between the host build environment, and the OE container.
2. reworks the build-nilrt Dockerfiles to work with pyrex, and adds a new script to create the containers automatically.
3. rewrites the `ni-init-build-env` script to work with pyrex. Sourcing the script works the same way as before, but it will now standup the shim layer while it drops you into the build/ directory.

Some caveats:
1. The `--sstate-cache` and `--download-dir` init script options are still in-development. Let me if you have feedback on how they should work.
2. The build-nilrt container currently has to be built locally before you can use pyrex the first time. In the future, we can create an image pipeline that pushes it to a public registry. Then we can configure pyrex to pull from that registry.
3. I didn't add any content to the build-nilrt container, beyond what we get for free from the pyrex ubuntu-20.04-oe container. If we find that hardknott requires more dependencies, we can add them to the `:docker/build-nilrt.Dockerfile` image layer.

# Usage
```bash
bash ./docker/create-build-nilrt.sh
source ni-init-build-env
bitbake ...
```

# Testing
Works on my machine. :) Let me know if you have issues.

@ni/rtos


chaitu236:
toaster isn't working with this. Is it a known limitation?

amstewart:
@chaitu236 
It isn't a limitation known to me. I haven't used toaster in any real capacity before. How are you using it?

chaitu236:
> @chaitu236 It isn't a limitation known to me. I haven't used toaster in any real capacity before. How are you using it?

I created a python virtual env, sourced it, installed[ toaster requirements](https://www.yoctoproject.org/docs/current/toaster-manual/toaster-manual.html#toaster-load-packages) into it, built docker, `source oe-init-build-env`, started toaster (`source ../source/bitbake/bin/toaster start webport=0.0.0.0:10000`), and built a recipe (`bitbake opkg:do_fetch`) - the recipe built but toaster's web page doesn't reflect it.

amstewart:
Hmm... honestly, I don't know enough about toaster to understand help. Do you need to use toaster for your dev workflow, or is it just something you're playing around with?

chaitu236:
> Hmm... honestly, I don't know enough about toaster to understand help. Do you need to use toaster for your dev workflow, or is it just something you're playing around with?

I've started using it recently and found that it's a good tool to keep track of previous build failures, find dependencies, see which recipes got built, track build times, find failures at a glance, etc.

So its a nice to have thing but probably not important enough to gate this PR - I'll figure out a fix/workaround or disable pyrex on my local builds.

amstewart:
OK. If you figure out a way to work, we can try to upstream it.

chaitu236 (approved):

PR #85, opened 2021-10-27
Support the Pyrex-enabled AZDO build pipeline

amstewart:
This PR enables NI-owned AZDO build machines, which are Pyrex-aware, to easily execute common build workflows within the NILRT source. In general it:
* Adds pipeline scripts to `:scripts/pipelines/` which automate the process of building the core and ~extras~ desirable packagegroups. These scripts are useful to have in the nilrt.git code to (a) reduce the code-size of our AZDO components and (b) allow for external builders and devs to build these artifacts using the same workflows as the build machines.
* Adds a new `:scripts/azdo/conf/pyrex.ini` file - specifically for the build machines. Though I tried to avoid having to maintain two `pyrex.ini` files in the source, there are just too many legitimate configuration differences in the AZDO build environment to warrant `sed`ing the top-level `pyrex.ini`.
    * Generally, the AZDO pyrex.ini takes more variables from the environment; such as the Pyrex container name and tag. It's possible that, once the new Pyrex piipeline is mature and building, that we decide we can converge the dev and AZDO INI files into a single file. I would be in favor of that.
* Modifies the existing `ni-oe-init-build-env` scripts to let the build machines pass through more variables (like `BUILDNAME` to the bitbake environment). I expect there will be a few more variables that need to be passed through, but they are not critical to getting the pipeline up and running at the moment.

## Testing
I have been using this branch as my `nilrt.git` source while developing the AZDO-half of the Pyrex-aware pipeline.

@ni/rtos


scripts/pipelines/build.test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+SCRIPT_ROOT=$(realpath $(dirname $BASH_SOURCE))
+
+. "${SCRIPT_ROOT}/build.common.sh" $@
+
+echo "INFO: Building test packages..."
+bitbake base-files
```
chaitu236, line 8: I assume this is for building a subset of packages to verify if bitbake is working and not for building packages that contain tests?

chaitu236 (approved):

scripts/pipelines/build.test.sh
amstewart: Yeah; correct. I just wanted a minimal pipeline to check that bitbake would actually run. I figure it is useful-enough to others that I added it to the git sources.

scripts/pipelines/build.common.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -e
+
+SCRIPT_ROOT=$(realpath $(dirname $BASH_SOURCE))
+
+. "${SCRIPT_ROOT}/../../ni-oe-init-build-env" $@
+
+bitbake --version
+#printenv | grep -i ^PYREX
```
shruthi-ravi, line 9: Is this required?

scripts/pipelines/delete-duplicate-ipks.sh
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env bash
```
shruthi-ravi, line 1: Not reviewing, assuming this was copied as is from azdo.

scripts/pipelines/build.extra-feeds.sh
```diff
@@ -46,7 +46,7 @@ core_feed_path="$(realpath ${positionals[0]})"
 echo "INFO: Building the extra package feed."
 set -x
 bitbake packagegroup-ni-desirable
-bitbake --continue packagegroup-ni-extra
+#bitbake --continue packagegroup-ni-extra
```
shruthi-ravi, line 49: Why not NOT run the build.extra-feeds script in its entirety?

shruthi-ravi (approved):

scripts/pipelines/build.extra-feeds.sh
shruthi-ravi: Nvm. I see that this also builds `desirable`.

scripts/pipelines/delete-duplicate-ipks.sh
amstewart: Correct; it was copied without modification.

scripts/pipelines/build.common.sh
amstewart: No; it was helpful to use during debug. I already have a local commit which adds a more formal function to log environment variables for debugging, which will replace this comment.

PR #87, opened 2021-11-04
Update instructions to compile the kernel

shruthi-ravi:
Split the existing instructions to compile and install the kernel
into separate sets for ARM and x64 targets. Update the instructions
to build NI drivers for both platforms based on the more recently
supported dkms framework.

@ni/rtos

### DIFF
--- a/KERNEL_SOURCE.txt
+++ b/KERNEL_SOURCE.txt
@@ -1,43 +1,39 @@
 ﻿NI Linux Real-Time Kernel Source README
 
 I. Introduction
-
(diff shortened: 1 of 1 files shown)


gratian (approved):

shruthi-ravi:
I've combined the instructions again, because the dkms portion works on both platforms as is, and so the difference between the two sets turned out to be minimal.

amstewart (approved):

gratian (approved):

PR #90, opened 2021-11-19
ni-oe-init-build-env: Check out submodules if we haven't already

bstreiff:
This is in the README, but this step is easy to forget to do. Make the script do it, if needed.

### DIFF
--- a/ni-oe-init-build-env
+++ b/ni-oe-init-build-env
@@ -145,6 +145,11 @@ pyrex_cleanup() {
 	unset PYREX_OEINIT PYREX_ROOT PYREX_TEMP_ENV_FILE pyrex_cleanup
 }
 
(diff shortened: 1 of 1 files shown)


amstewart:
I don't think its necessary to check specifically for the pyrex submodule. Users should expect that if they change branches, the submodule content of the repo may shift, and they may need to reinitialize. Other submodules, like the `meta-orgconf` have dropped from the submodules without being mentioned by the init script, and likewise layers like `meta-rauc` were added without this kind of accounting.

amstewart:
@bstreiff Are you going to push back on my NACK? Or do you want to drop this PR?

bstreiff:
Drop; I don't have a solution that I think addresses your concerns.

PR #91, opened 2021-11-19
Refresh the buildVM script and use Pyrex

amstewart:
The AZDO pipeline build machines do not natively include ISO and VM utilities like `genisoimage`, `qemu-utils`, and the QEMU system binaries. So they cannot natively run the `buildVM.sh` script. Since they are already using the Pyrex container, we may as well just add these utilities to that container context.

And since I already need to touch the `buildVM` script to make that work, I decided to just make it a first class scriptlet in the `scripts/pipelines` directory. This PR moves the `buildVM` script into a `build.vms.sh` scriptlet, and majorly refreshes it to be simpler and more modern. Most importantly, it now has sane defaults. Simply calling the script from anywhere in the repo, after the recovery media has been created, is sufficient to create a QEMU VM archive.

Otherwise, the major steps of the build-vm script haven't fundamentally changed.

# Testing
Tested many times on my dev machine.

@ni/rtos


amstewart:
@shruthi-ravi Let's hold off on merging this until I'm back from the holidays. I'm not 100% sure that the new Dockerfile won't break the AZDO pipeline until I push a new container manually.

scripts/pipelines/build.vms.sh
```diff
@@ -0,0 +1,261 @@
...
+
+    sed -i "s%\${PRIMARY_DISK}%${primary_disk}%g" "${script_destination_path}"
+    sed -i "s%\${VM_MACHINE_UUID}%${vm_uuid}%g" "${script_destination_path}"
+    sed -i "s%\${VM_DISK_UUID}%${disk_uuid}%g" "${script_destination_path}"
+    sed -i "s%\${VM_MEM_SIZE_MB}%${memory_mb}%g" "${script_destination_path}"
+}
+
+# Source the common build setup script, so that we're pyrex-enabled and our
+# working directory is changed to the build directory.
+. "${SCRIPT_ROOT}/build.common.sh"
```
shruthi-ravi, line 210: Does this work fine if this script is run in an already Pyrex-enabled environment? I haven't set up the Pyrex stuff yet, so I don't know.

scripts/pipelines/build.vms.sh
```diff
@@ -0,0 +1,261 @@
...
+    1024 MB.
+-n,--name NAME
+    The desired name for the VM archive.
+-r,--recovery-iso RECOVERY_ISO
+	The filepath to the NILRT recovery media ISO, which will provision the
+    primary disk.
+
+# Environmentals
+MACHINE
+	The machine architecture of the created VM. Only the 'x64' architecture is
```
shruthi-ravi: How do you control the value of this variable? i.e. restrict it to only x64? Why not have it as a constant for now? Then, we don't have to do any validation on the value.

scripts/pipelines/build.vms.sh
```diff
@@ -0,0 +1,261 @@
...
+-d,--disk-size DISK_SIZE
+    The primary size of the VM's main storage device. Defaults to 4096 MB.
+-i,--images-dir IMAGES_DIR
+	The filepath to the bitbake deploy/images directory which contains the
+    recovery media ISO.
+-m,--memory MEMORY_MB
+	The desired size (in megabytes) for the VM's virtual RAM. Defaults to
+    1024 MB.
+-n,--name NAME
+    The desired name for the VM archive.
```
shruthi-ravi: Defaults to "nilrt".

scripts/pipelines/build.vms.sh
```diff
@@ -0,0 +1,261 @@
...
+	local disk_uuid=""
+
+	install \
+		--mode=0755 \
+		"${SCRIPT_RESOURCE_DIR}/${script_template}" \
+		"./${script_destination_path}"
+
+    sed -i "s%\${PRIMARY_DISK}%${primary_disk}%g" "${script_destination_path}"
+    sed -i "s%\${VM_MACHINE_UUID}%${vm_uuid}%g" "${script_destination_path}"
+    sed -i "s%\${VM_DISK_UUID}%${disk_uuid}%g" "${script_destination_path}"
```
shruthi-ravi: These variables seem to be used only in the old .vbox templates.

shruthi-ravi (approved):

billpittman (approved):

scripts/pipelines/build.vms.sh
amstewart: Yes. If you look at the other files in `:scripts/pipelines/`, they all start by sourcing the `build.common.sh` script. Pyrex initialization doesn't drop the build machine into a persistent container context, it just sets up the build/ directory in a special way. So you can arbitrarily run through the "initialization" codepaths afterward, and they will still execute in the host machine context. It's only when you run one of the pyrex-container commands that the container is started.

scripts/pipelines/build.vms.sh
amstewart: I don't restrict it anywhere. We've used the `MACHINE` variable in so many NILRT scripts that I was thinking of it as a *standard* switch that NILRT builders will know how to use. And I continued to use it here in the case that we add more architecture support in the future.

But now that you have me thinking about it again - those aren't good reasons. If we add more arch-support in the future, we will need to refactor anyway. I'll drop the variability and hardcode the `x64` value for now, for simplicity.

scripts/pipelines/build.vms.sh
amstewart: Will fix.

scripts/pipelines/build.vms.sh
amstewart: True. I'll remove them.

amstewart:
# Patch V2
* Addressed @shruthi-ravi's feedback.
    * Removed unused UUID template vars.
    * Updated usage text.
    * Hardcoded the `x64` MACHINE value.

shruthi-ravi (approved):

PR #92, opened 2021-12-01
build.vms: remove remaining uuid references

amstewart:
The `vm_uuid` and `disk_uuid` variables are obsolete and place a requirement on the build host to have the `uuidgen` binary. Remove them.

# Testing
* With this patchset, I can now build the virtual machines using the relatively minimal nicentral container in my dev machine. The rfmibuild machines should be similar.

@ni/rtos

### DIFF
--- a/scripts/pipelines/build.vms.sh
+++ b/scripts/pipelines/build.vms.sh
@@ -192,9 +192,6 @@ write_vm_startup_script() {
 	local script_destination_path=$2
 	local primary_disk=$3
 
-	local vm_uuid=$(uuidgen)
-	local disk_uuid=""
-
 	install \
 		--mode=0755 \
 		"${SCRIPT_RESOURCE_DIR}/${script_template}" \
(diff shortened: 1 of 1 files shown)


scripts/pipelines/build.vms.sh
```diff
@@ -192,9 +192,6 @@ write_vm_startup_script() {
 	local script_destination_path=$2
 	local primary_disk=$3
 
-	local vm_uuid=$(uuidgen)
```
shruthi-ravi, line 195: Is there a Dockerfile (additionally installing this binary) that also has to be updated?

shruthi-ravi (approved):

scripts/pipelines/build.vms.sh
amstewart: There is not. I found this issue when running the script on my xidax machine - which doesn't have the binary. Part of the wrongness of this line is that it doesn't use the `$PYREX_RUN` wrapper, and is trying to run the command on the host machine. So it breaks on any machine without the binary.

PR #97, opened 2021-12-09
README: update README with pyrex information

amstewart:
Update the top-level project README with information about how to build
and use the new pyrex-enabled init script.

Also remove most of the rest of the README, since it is outdated
information.

We can add back steps for building the feeds and image targets as they solidify for hardknott.

@ni/rtos 

@shruthi-ravi I'm *pretty sure* this is all you need to do. Let me know if I forgot a step.


amstewart:
# Patch V2
* Added a forgotten step to create the pyrex container image.

shruthi-ravi (approved):
Thanks. I've been able to use these instructions to build packages.

amstewart:
# Patch V3
* Squashed fixup commits. No substantive changes.

PR #106, opened 2022-02-03
Update Hardknott README.md

ni-jxie:
Add more details to Hardknott README.md to describe how to build a package/package group/images.

@ni/rtos This is how I understand the process so far. Feel free to correct/add anything you think that is missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,73 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
     bitbake --version  # If this succeeds, you're done.
     ```
 
-    <font color=lightgreen>[NI]</font> builders who are connected to the NI  corporate network should specify `-org` in their init script args, to provoke the script into adding the `ni-org.conf` snippet to your bitbake directory. External builders *should not* use `--org`.
+    <font color=lightgreen>[NI]</font> builders who are connected to the NI corporate network should specify `-org` in their init script args, to provoke the script into adding the `ni-org.conf` snippet to your bitbake directory. External builders *should not* use `--org`.
+
+5. Build the package or packages that you want for your target. For example, to build Python, Ruby, and Apache for x64 targets, run the following commands:
+
+        bitbake python ruby apache2
+
+    To build all supported OpenEmbedded packages in NI's feed, run the following commands to build these packagegroups:
+
+        bitbake packagefeed-ni-core
+        bitbake --continue packagefeed-ni-extra
+
+    **NOTE** If a package within a package group is updated, rebuilding the package group will automatically rebuild that package and all package depending on that package.
+
+    **NOTE** The configuration files (build/conf/*.conf) can optionally be changed to reflect the desired build settings instead of setting environment variables.
+
+    **NOTE** Building packages through OpenEmbedded can use significant disk space, on the order of 100s of gigabytes. If you are preparing a virtual machine to build images, make sure to allocate sufficient disk space.
+
+    The resulting ipk files that can be installed through opkg exist at the following directory:
+
+        tmp-glibc/deploy/ipk/...
+        
+6. Bitbake can transform tmp-glibc/deploy/ipk/<tune> into package feeds when you run the following command:
+
+        bitbake package-index
+
+    You must rebuild the package-index before building images or if any package whom the images you are trying to build depends on has been changed and rebuilt.
+    
+7. Building various images
+
+    **NOTE** You must build packagefeed-ni-core and package_index first to build images.
+
+    * Build a safemode image by running the following command:
+
+            bitbake nilrt-safemode-rootfs
+
+        The resulting root file system images for the NILRT safemode image is located at the following paths:
+
+            tmp-glibc/deploy/images/x64/nilrt-safemode-rootfs-x64.tar.gz
+    
+        You can install this on target by copying the file over to the target and running the following command:
+
+            tar xf nilrt-safemode-rootfs-x64.tar.gz -C /boot/.safe/
+    
+    * Build a runmode image by running the following command:
+
+            bitbake nilrt-base-system-image
+
+        The resulting root file system images for the NILRT runmode image is located at the following paths:
+
+            tmp-glibc/deploy/images/x64/nilrt-base-system-image-x64.tar.gz
+    
+        You can install this on target by copying the file over to the target while the target is in safe mode and running the following commands:
+
+            tar xf nilrt-base-system-image-x64.tar.gz
+            tar xf data.tar.gz -C /mnt/userfs && ./postinst
+
+    * Build a bootable recovery media by running the following command:
+
+            bitbake nilrt-recovery-media
+
+        The bootable recovery media, which you can install onto a USB memory stick or burn to a CD, is located at the following path:
+
+            tmp-glibc/deploy/images/x64/nilrt-recovery-media-x64.iso
+
+        Boot your NI Linux Real-Time compatible hardware from the recovery media and follow on-screen instructions to perform a factory reset.
+
+    **NOTE** By default, National Instruments software is pulled from a feed hosted on ni.com. You can redirect to a mirror by setting IPK_NI_SUBFEED_URI to any URI supported by opkg in your org.conf,site.conf, or auto.conf.
 
 ---
 


README.md
```diff
@@ -49,6 +49,59 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
...
+    
+7. Building various images
+
+    Build a safemode image by running the following commands:
+
+        bitbake nilrt-safemode-rootfs
+    
+    Build a runmode image by running the following commands:
+
+        bitbake nilrt-runmode-rootfs
```
chaitu236: Better specify `nilrt-base-system-image` instead of `nilrt-runmode-rootfs` as the output of `nilrt-base-system-image` can directly be used from MAX.

README.md
```diff
@@ -49,6 +49,59 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
...
+
+    Build a bootable recovery media by running the following commands:
+
+        bitbake nilrt-recovery-media
+
+    **NOTE** You must build everything in packagefeed-ni-core(step 5) and build package_index (step 6) to build images.
```
chaitu236: This isn't strictly true as `packagefeed-ni-extra` is not necessary for building images - only `packagefeed-ni-core` is required.

README.md
```diff
@@ -49,6 +49,59 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
...
+        bitbake python ruby apache2
+
+    To build every package in NI's feed, run the following commands:
+
+        bitbake packagefeed-ni-core
+        bitbake --continue packagefeed-ni-extra
+
+    **NOTE** The configuration files (build/conf/*.conf) can optionally be changed to reflect the desired build settings instead of setting environment variables.
+
+    **NOTE** Building packages through OpenEmbedded can use significantdisk space, on the order of tens of gigabytes. If you are preparing a virtual machine to build images, make sure to allocate sufficient disk space.
```
chaitu236: significantdisk -> significant disk

chaitu236 (approved):

README.md
```diff
@@ -49,6 +49,59 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
 
     <font color=lightgreen>[NI]</font> builders who are connected to the NI  corporate network should specify `-org` in their init script args, to provoke the script into adding the `ni-org.conf` snippet to your bitbake directory. External builders *should not* use `--org`.
 
+5. Build the package or packages that you want for your target. For example, to build Python, Ruby, and Apache for Zynq targets, run the following commands:
```
shruthi-ravi: Change to x64 targets. Only x64 is supported in hardknott.

README.md
```diff
@@ -49,6 +49,59 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
...
+
+        bitbake python ruby apache2
+
+    To build every package in NI's feed, run the following commands:
```
shruthi-ravi: I think `every package in NI's feed` is misleading. We mean only OE packages, right? Maybe something like `to build all supported OpenEmbedded packages`?

README.md
```diff
@@ -49,6 +49,59 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
...
+    **NOTE** The configuration files (build/conf/*.conf) can optionally be changed to reflect the desired build settings instead of setting environment variables.
+
+    **NOTE** Building packages through OpenEmbedded can use significantdisk space, on the order of tens of gigabytes. If you are preparing a virtual machine to build images, make sure to allocate sufficient disk space.
+
+    The resulting ipk files that can be installed through opkg exist at
+the following directory:
+
+        tmp-glibc/deploy/ipk/...
+        
+6. Bitbake can transform tmp-glibc/deploy/ipk/ into a package feed when you run the following command:
```
shruthi-ravi: Do you need to say something like:
`Bitbake can transform tmp-glibc/deploy/x64/<tune> into package feeds when....`
to call out the multiple tune-dependent package feeds that will be created?

README.md
shruthi-ravi: I realized after reviewing that this text has been mostly copied from the sumo README. So I'll leave it to you, if you want to keep this README as is, or make this and the following updates.

shruthi-ravi (approved):
Thanks for making the updates! I have some minor comments below.

README.md
ni-jxie: Do we advertise to external clients out of NI that they can go ahead and do that from MAX? Since this is external facing, I don't want to rely on the internal directory structure of MAX.

README.md
ni-jxie: Yeah, I meant to say packagefeed-ni-core is the only one needed. Let me reword this a bit.

README.md
chaitu236: We don't but I don't see how they can use the output of `nilrt-runmode-rootfs` directly either.

Maybe we can mention that the .tar file that gets built from `nilrt-base-system-image` can be copied to a target in safemode, unpacked, then `tar xf data.tar.gz -C /mnt/userfs && ./postinst` can be run in lieu of provisioning from MAX?

I don't have a strong opinion here, so feel free to drop if you wish.

README.md
ni-jxie: On my system, there 3 directories under `ipk` are
```
./tmp-glibc/deploy/ipk/
├── all
├── core2-64
└── x64
```

And they all have ipks in them. Are you saying that only the ones in x64 will be sucked into the feed?

README.md
ni-jxie: Fixed.

README.md
ni-jxie: Fixed.

README.md
ni-jxie: Fixed.

README.md
ni-jxie: Updated.

README.md
ni-jxie: If they cannot make use of the output of `nilrt-runmode-rootfs` directly, then what about "nilrt-safemode-rootfs"? If we want it to be more useful for customers, instead of rootfs, should I refer to `nilrt-base-system-image` instead and omit the rootfs? Would that make more sense?

chaitu236: We want to point customers to the recipes that produce artifacts that are most useful.

In case of runmode, I see `nilrt-base-system-image` as a better choice since it also includes the `postinst` script which is required for successful provisioning (`nilrt-runmode-rootfs` doesn't include this).

In case of safemode, although customers may find `nilrt-recovery-media` to be a better choice to upgrade safemode, I think its ok to mention `nilrt-safemode-rootfs` also as it contains everything needed to upgrade an existing safemode installation (by unpacking the output archive and copying file from it into /boot/.safe/). So I'd prefer to keep `nilrt-safemode-rootfs` like it is now.

README.md
```diff
@@ -47,7 +47,59 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
...
+        bitbake python ruby apache2
+
+    To build all supported OpenEmbedded package in NI's feed, run the following commands:
+
+        bitbake packagefeed-ni-core
+        bitbake --continue packagefeed-ni-extra
+
+    **NOTE** The configuration files (build/conf/*.conf) can optionally be changed to reflect the desired build settings instead of setting environment variables.
+
+    **NOTE** Building packages through OpenEmbedded can use significant disk space, on the order of tens of gigabytes. If you are preparing a virtual machine to build images, make sure to allocate sufficient disk space.
```
jeminor: We're in the 100s of G now by the way.

README.md
jeminor: I'm ok with listing nilrt-base-system-image, provided that we clearly indicate that using so "voids the warranty" on support.  That ok with everybody?

shruthi-ravi: Why does using the nilrt-base-system-image void support? I don't follow. I also prefer to have instructions to build nilrt-base-system-image instead of the rootfs.

README.md
shruthi-ravi: No, I mean that a feed will be created for each directory you've listed. For example, 
image
which shows three separate feeds, one for each compiler "tune", identified by the Packages.gz file. You'll have to register each of those as a separate feed with opkg. Check out the contents of `/etc/opkg/base-feeds.conf` on a target.
Does that make sense?

ni-jxie: Yeah, got it. I got what you meant now.

README.md
ni-jxie: Updated to nilrt-base-system-image.

@jeminor What is the line between "voiding the warranty" and valid use case? If they just built our branch as it is without making changes, I am assuming that is a valid use case, is that correct? Only if they change the recipes or configuration materially. Then it is "voiding the warranty"?

ni-jxie: As discussed, the MIT license and the NI EULA already covered the use cases of client building from source.

jeminor (approved):

README.md
```diff
@@ -47,7 +47,73 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
...
+    
+        You can install this on target by copying the file over to the target and running the following command:
+
+            tar xf nilrt-safemode-rootfs-x64.tar.gz -C /boot/.safe/
+    
+    * Build a runmode image by running the following command:
+
+            bitbake nilrt-base-system-image
+
+        The resulting root file system images for the NILRT safe-mode image is located at the following paths:
```
shruthi-ravi: for the NILRT `runmode` image?

README.md
```diff
@@ -47,7 +47,73 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
...
     ```
 
-    <font color=lightgreen>[NI]</font> builders who are connected to the NI  corporate network should specify `-org` in their init script args, to provoke the script into adding the `ni-org.conf` snippet to your bitbake directory. External builders *should not* use `--org`.
+    <font color=lightgreen>[NI]</font> builders who are connected to the NI corporate network should specify `-org` in their init script args, to provoke the script into adding the `ni-org.conf` snippet to your bitbake directory. External builders *should not* use `--org`.
+
+5. Build the package or packages that you want for your target. For example, to build Python, Ruby, and Apache for x64 targets, run the following commands:
+
+        bitbake python ruby apache2
+
+    To build all supported OpenEmbedded package in NI's feed, run the following commands to built these packagegroups:
```
shruthi-ravi: NIT: s/package/packages, s/built/build

shruthi-ravi (approved):

README.md
ni-jxie: Good catch.

README.md
ni-jxie: Done.

PR #110, opened 2022-02-16
ni-org.conf: disable versionator

amstewart:
Temporarily stop using the 'versionator' PR server, while it is
experiencing downtime.

---

I'll revert this PR, once versionator is back online.

# Testing

Ran `bitbake -e glibc`, which would hang if the pr-server doesn't respond.

@ni/rtos


chaitu236:
https://wiki.yoctoproject.org/wiki/PR_Service says

> It should be noted that the PR information is not maintained as part of the sstate packages. If you maintain a sstate feed, its expected that either you run builders contributing to the sstate feed with a shared PR service, or you don't run one. Mixing the two will lead to problems.

Does this mean we should also disable sstate cache?

amstewart:
I don't know what "problems" the wiki is referring to, and I can't think of any particular reason that it would break the sstate cache.

shruthi-ravi (approved):

PR #141, opened 2022-05-11
dist-feed: Update requirements to new pypi location.

jeminor:
Update requirements to new pypi location to fix dist build breakage.

Signed-off-by: James Minor <james.minor@ni.com>

### DIFF
--- a/scripts/jenkins/dist-feed/requirements.txt
+++ b/scripts/jenkins/dist-feed/requirements.txt
@@ -1,5 +1,4 @@
 Jinja2>=2.10.1
 PyYAML>=5.1.2
---trusted-host ni-pypi.amer.corp.natinst.com
---index-url http://ni-pypi.amer.corp.natinst.com/simple/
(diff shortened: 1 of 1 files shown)


scripts/jenkins/dist-feed/requirements.txt
```diff
@@ -1,5 +1,4 @@
 Jinja2>=2.10.1
 PyYAML>=5.1.2
---trusted-host ni-pypi.amer.corp.natinst.com
```
amstewart, line 3: What's the reason that we can remove this line?

shruthi-ravi (approved):

scripts/jenkins/dist-feed/requirements.txt
jeminor: The new URI is https, thus doesn't have to be explicitly trusted.

amstewart (approved):

PR #144, opened 2022-06-01
pyrex: enable containers-by-digest

amstewart:
The current pyrex implementation assumes that users wish to start
containers using the `$name:$tag` format. Users may not use the
`$name@$digest` format, which is preferable in environments where
build-reproducibility is required.

Change the pyrex.ini to eschew this assumption and update the init
script to allow the user to pull containers by digest.

----

This PR is needed to use static containers in our AZDO pipeline.

# Testing
* Exported a `PYREX_BUILD_NILRT_IMAGE` variable with the full digest name of the latest `hardknott`-tagged `build-nilrt` container on `rnd-builds.repos.natinst.com`, and ran the init script. It correctly used the container-by-digest.
* Unexported the image variable and ran the initscript. Pyrex started using the `build-nilrt:latest` tag from my dev-machine.

### DIFF
--- a/ni-oe-init-build-env
+++ b/ni-oe-init-build-env
@@ -133,6 +133,9 @@ export PYREX_CONFIG_BIND="${NILRT_ROOT}"
 export PYREX_OEINIT="${NILRT_ROOT}/sources/openembedded-core/oe-init-build-env"
 export PYREX_ROOT="${NILRT_ROOT}/sources/pyrex"
 
+export PYREX_BUILD_NILRT_IMAGE="${PYREX_BUILD_NILRT_IMAGE:-build-nilrt:latest}"
+echo "INFO: Using pyrex image: ${PYREX_BUILD_NILRT_IMAGE}"
+
 
 unset SCRIPT_ROOT
 
--- a/pyrex.ini
+++ b/pyrex.ini
@@ -24,23 +24,14 @@ confversion = 2
 # The Container engine executable (e.g. docker, podman) to use. If the path
 # does not start with a "/", the $PATH environment variable will be searched
 # (i.e. execvp rules)
-#engine = docker
+engine = docker
 
 # The type of image to build
 #imagetype = oe
 
-# As a convenience, the name of a Pyrex provided image
-# can be specified here
-#image = build-nilrt
-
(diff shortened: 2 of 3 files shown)


amstewart:
# Patch V2
* commit spelling.

shruthi-ravi (approved):
Do you want to hold this change until the build is stable? I suppose there is an associated azdo change too that needs to go in.

ni-oe-init-build-env
```diff
@@ -133,6 +133,9 @@ export PYREX_CONFIG_BIND="${NILRT_ROOT}"
 export PYREX_OEINIT="${NILRT_ROOT}/sources/openembedded-core/oe-init-build-env"
 export PYREX_ROOT="${NILRT_ROOT}/sources/pyrex"
 
+export PYREX_BUILD_NILRT_IMAGE="${PYREX_BUILD_NILRT_IMAGE:-build-nilrt:latest}"
```
shruthi-ravi, line 136: Once a release branch is created, should the default be updated to the respective release-tagged container, like what you are doing in the sumo changelist.

shruthi-ravi: Or why not use the same logic used for the sumo branch?

amstewart: I have bigger plans for setting and using the codename in hardknott, as well as more changes to simplify the pyrex configuration. But I wanted to get these MVP changes merged first, so that I can unblock the work to switch to using static container IDs in AZDO.

amstewart:
> Do you want to hold this change until the build is stable? I suppose there is an associated azdo change too that needs to go in.

I'm OK putting it in now. I believe the only remaining current issue with the pipelines affects only the `nilrt_os_common` component, which won't be affected by these changes. And getting these in will unblock my AZDO work and let me start PR validation runs.

shruthi-ravi:
> I'm OK putting it in now. I believe the only remaining current issue with the pipelines affects only the nilrt_os_common component, which won't be affected by these changes. And getting these in will unblock my AZDO work and let me start PR validation runs.

I'm less confident with pulling additional changes while the build is already down :). But I'll leave the decision to you.

PR #150, opened 2022-06-08
Update packagegroup to build extra feed correctly

shruthi-ravi:
The sumo extras feed was not being built completely because
packagefeed-ni-extra is a hardknott-specific packagegroup name. Fix
the feed by updating to sumo packagegroup `packagegroup-ni-extra`.

Haven't tested the change, I expect it to just work.


amstewart (approved):
I would've sworn that I accounted for this when I backported this script from the hardknott mainline. I guess not. :(

PR #162, opened 2022-06-30
README: Add section to set ni-main feed version

chaitu236:
Also fixed list numbering to take advantange of github markdown's
auto-numbering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,12 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
 
     **NI builders** who are connected to the NI corporate network should specify `-org` in their init script args, to provoke the script into adding the `ni-org.conf` snippet to your bitbake directory. External builders *should not* use `--org`.
 
+   If you are building on a `nilrt/master/*` branch ref (rather than a release branch) **and** if you are building outside of the NI corporate network, you will need to set the version of the `ni-main` opkg feed to one which has already been published to `download.ni.com`. Do this by setting the `NILRT_MAIN_FEED_VERSION` bitbake variable to the latest published release. eg.
+
+   ```
+   echo 'NILRT_MAIN_FEED_VERSION = "2022Q3"' >> ./conf/local.conf
+   ```
+
 5. #### Build package or packagegroups
    For example, to build Python, Ruby, and Apache for x64 targets, run the following commands:
 


README.md
```diff
@@ -31,10 +31,10 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
     git submodule update --remote --checkout
     ```
 
-2. #### Install docker
+1. #### Install docker
```
amstewart: I would normally agree with you that we should use auto-numbering; but there is a significant demographic of developers who will read this README in a terminal. So I would prefer to keep the numbering explicit in the markdown, even though it is a little more work.

README.md
```diff
@@ -43,7 +43,13 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
     docker images build-nilrt  # should print the image you just built
     ```
 
-4. #### Set up build environment
+1. #### Set `ni-main` feed version
+   If you're building a pre-release branch (`nilrt/master/hardknott`), set `NILRT_MAIN_FEED_VERSION` in `nilrt/sources/meta-nilrt/conf/distro/nilrt.inc` to last released ni-main feed.
```
amstewart: The `nilrt/master/hardknott` ref is going to become obsolete once we rebase. And I think builders should be overwriting the default variable value in their `local.conf`, rather than modifying the meta-nilrt layer. What about...

"""
If you are building on a `nilrt/master/*` branch ref (rather than a release branch) **and** if you are building outside of the NI corporate network, you will need to set the version of the `ni-main` opkg feed to one which has already been published to [`download.ni.com`](http://download.ni.com/#ni-linux-rt/feeds/). Do this by setting the `NILRT_MAIN_FEED_VERSION` bitbake variable to the latest published release. eg.
```bash
echo 'NILRT_MAIN_FEED_VERSION = "2022Q3"' >>./conf/local.conf
```
"""

And then we should probably put this task into the "Set up the build environment" step.

amstewart (changes requested):

README.md
chaitu236: Done

README.md
chaitu236: Done

shruthi-ravi (approved):

amstewart (approved):

PR #164, opened 2022-07-08
Add a digest of the package feed differences between NILRT 9.0 and NILRT 8.11

dylan-matthew-turner:
# Justification
 
There is a desire to see the packages dropped, added, and changed from one version of NILRT to the next, specifically in the core/ and extra/ feeds.

There are several of these changes for each version, so ideally it needs to be automated and not clutter CHANGELOG.md

# Implementation

- Create a script to pull package information from the core/ and extra/ feeds for two different versions of NILRT and generate a report of the differences
    + The script inserts a new report for version X to version Y under the header part of the created "FEED_CHANGELOG.md"
    + Example for generating 8.11 to 9.0:
        ```
        ./scripts/diff-pkg-changes.py \
            nilrt/21.8/sumo 8.11 nilrt/22.5/hardknott 9.0 \
            --url http://nickdanger.amer.corp.natinst.com/feeds
        ```
    + Note: the script can be adjusted to care about more changes like changes to "Replaces" or changes to the MD5Sum and things like that. Currently it cares about:
        * Version
        * Depends
        * Replaces
        * Section
        * Architecture
        * Size
- Generate feed change report for 8.11 to 9.0 to FEED_CHANGELOG.md via added script
- Add note about FEED_CHANGELOG.md (and link) as well as links to each feed subsection in CHANGELOG.md


dylan-matthew-turner:
Forgot to note added packages in script. I will add shortly

dylan-matthew-turner:
Missed a type annotation. I'll update that as well for consistency's sake

shruthi-ravi:
Hey @dylan-matthew-turner. Can you ensure that your commits are squashed into functional changes? This is a little different from how we do things in azdo. When the maintainer merges your changes in GitHub, he/she will do a rebase of your changes on top of the master/hardknott branch and not a squash commit. So you need to ensure that your commit and commit messages have been formatted accordingly. Take a look at the history of commits in the nilrt/master/hardknott [branch](https://github.com/ni/nilrt/commits/nilrt/master/hardknott/README.md) for a reference point. You can do a force push once you have squashed your commits.
Also, ensure that your commit message header is in this format:
<recipe/file name>: short description of change

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,291 @@
...
+        # Then construct this class
+        self.old_branch: str = args.old
+        self.new_branch: str = args.new
+        self.feed_base: str = args.url
+        self.old_vers: str = args.old_vers
+        self.new_vers: str = args.new_vers
+
+# A structure to hold the info about a package
+@dataclass
+class Package:
```
amstewart, line 95: The [opkg-utils](https://git.yoctoproject.org/opkg-utils/tree/opkg.py) repo in the yocto project contains some useful opkg classes and functions that we can/should reuse here, if we can.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,291 @@
...
+        return diff_str
+
+# Get the year/identity of each branch's feed by pulling the info from a file on GitHub
+def get_feed(branch_name: str) -> str:
+    # Get the file with the info
+    url: str = GITHUB_BASE + '/' + branch_name + '/' + GITHUB_CONF_FNAME
+    request: Response = get(url)
+    if request.status_code != 200:
+        print('Failed to find branch %s!' % branch_name)
+        exit()
```
amstewart: I think `exit()` is only intended to be used within the interactive python interpreter. Can you switch this to propagate an error upwards instead?

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,291 @@
...
+    feed_url_base: str, feed_version: str, feed_name: str, branch_name: str
+) -> list['Package']:
+    pkgs: list['Package'] = []
+
+    # Get the package file
+    url: str = feed_url_base + '/' + feed_version + '/x64/' + feed_name + '/all/Packages'
+    request: Response = get(url)
+    if request.status_code != 200:
+        print('Failed to download feed for %s!' % branch_name)
+        exit()
```
amstewart: Same comment about `exit()`.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,291 @@
...
+def update_changelog(report: str) -> None:
+    old_log: str = ''
+    with open('FEED_CHANGELOG.md', 'r') as changelog:
+        old_log = changelog.read()
+    new_log = FEED_CHANGELOG_START + report + old_log.replace(FEED_CHANGELOG_START, '')
+    with open('FEED_CHANGELOG.md', 'w') as changelog:
+        changelog.write(new_log)
+
+if __name__ == '__main__':
+    main()
```
amstewart: There are a couple places in the module (in the feed Request logic) that can logically fail without returning a python error. Could you handle these cases and have the script return a sane error code with `sys.exit(1)`?

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,291 @@
...
+    return report
+
+# Update the FEED_CHANGELOG with the new version info
+def update_changelog(report: str) -> None:
+    old_log: str = ''
+    with open('FEED_CHANGELOG.md', 'r') as changelog:
+        old_log = changelog.read()
+    new_log = FEED_CHANGELOG_START + report + old_log.replace(FEED_CHANGELOG_START, '')
+    with open('FEED_CHANGELOG.md', 'w') as changelog:
+        changelog.write(new_log)
```
amstewart: What do you think about using a more UNIX-y output design? I'm thinking that we could remove the hardcoded `FEED_CHANGELOG.md` paths, and instead have this script write the `report` as a JSON structure to stdout. Then we can keep a log of all the changes in the nilrt.git source like
```bash
python3 diff-pkg-changes.py ... >>./docs/feed-changelog.json
```

Then we can have a small companion script that reads json from stdin and writes out the markdown changelog like
```bash
python3 emit-feed-changelog-markdown.py <./docs/feed-changelog.json >./docs/feed-changelog.md
```

I would use `jinja2` templating for generating the markdown, but an implementation similar to what you are already doing in this script is probably fine too.

What do you think?

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,291 @@
...
+    '# Changes to NI Linux RT Package Feeds\n\n' \
+        + '## Description\n\n' \
+        + 'This documents the full set of changes to individual packages ' \
+        + 'in the core/ and extra/ feeds for each version release.\n\n' \
+        + 'The changes are too long to keep in the [Changelog](/CHANGELOG.md) itself, ' \
+        + 'so they are kept here\n\n'
+
+def main() -> None:
+    # Get user input
+    settings: 'Settings' = Settings()
```
amstewart: I think you primarily broke out `Settings` into its own class in order to keep the code neat, but I might suggest a slightly different architecture that will keep the code more reusable.

```python3
class CLIParser(ArgumentParser):

    def __init__(self):
        super().__init__(self)
        self.add_argument(...
        ...

def main(args) -> int:
    old_branch: str = args.old_branch
    ...
    return 0

if __name__ == "__main__":
    parser = CLIParser()
    args = parser.parse_args()
    sys.exit(main(args))
```

That way, if someone wants to reuse this module from python, they can source the module and use the `CLIParser` to parse their own argument values (or populate the `args` namespace manually), and then call the `main()` method - all from within python.

FEED_CHANGELOG.md
```diff
@@ -0,0 +1,2388 @@
...
+
+## Description
+
+This documents the full set of changes to individual packages in the core/ and extra/ feeds for each version release.
+
+The changes are too long to keep in the [Changelog](/CHANGELOG.md) itself, so they are kept here
+
+## Changes from Version 8.11 to Version 9.0
+
+### Changes to core/
```
amstewart: I think this is probably a lot more info than users are looking for in a markdown file. I think they would prefer to have a more condensed digest of "what was added/removed". Something like...

```md
## Feed Changes from Version 8.11 to Version 9.0

### core/
* Added
    * mypaint-brushes-1.0
    * mypaint-brushes-1.0-dbg
    * mypaint-brushes-1.0-dev
* Removed
    * xorg-minimal-fonts
```

* Disregard package version bumps. There will be too many of them to be comprehensible.
* Disregard packages with the `-lic` suffix. They are just license subpackages, and readers probably won't care about them in the context of this document.
* Keep each entry as concise as possible.
* Remove information about sizes and dependencies changing. That data is useful, but not for this digested document, I don't think.
    * If we decide to follow my suggestion about splitting up the scripts that generate this file, and keeping a json log of all the raw *change* data, then future users can pull this more-detailed information from the json, if they are interested.

amstewart (changes requested):
This is a good place to start, and I have some architectural suggestions below.

Also, your commit summary lines are good; but could you add justifications and explanations to your commit bodies. The commits are the fundamental unit of change in yocto repos, so they need to carry their own information. The PR is just for code review. Also remember to sign them off with `-s`. :)

scripts/diff-pkg-changes.py
dylan-matthew-turner: I like this, but I have one question about the JSON: are we keeping just the most recent changes, or should it contain multiple sets of changes over time?

Because if we're updating the JSON overtime to keep changes every new version like the changelog, simply updating it with append (>>) won't work because when importing the JSON, python expects:
```
{
  "something": {...},
  "something_else": {...}
}
```
but append would create this:
```
{
  "something": {...}
}
{
  "something_else": {...}
}
```

There's a few ways we could go about creating the valid JSON like having an input/output file as an argument or modifying the output before appending in bash, what would be the preference here?

dylan-matthew-turner: Another option could be a "JSON Appender" script and we could pipe the JSON output into it and let it handle the appending like `./scripts/diff-pkg-changes.py ... | ./scripts/append-feed-changelog-json.py ./docs/feed-changelog.json > ./docs/feed-changelog.json`

EDIT: So the whole thing would work like this:
```
./scripts/diff-pkg-changes.py ... \
  | ./scripts/append-feed-changelog-json.py ./docs/feed-changelog.json \
  > ./docs/feed-changelog.json
./emit-feed-changelog-markdown.py ./docs/feed-changelog.json  > ./docs/feed-changelog.md
```

amstewart: We can just output the whole JSON object without newlines and then use the [jsonlines](https://jsonlines.org/) standard to store it. The changelog digest script would then just iterate through each line of the file and load it as a json object.

scripts/diff-pkg-changes.py
dylan-matthew-turner: I added an error corresponding to being able to find the file under said branch

scripts/diff-pkg-changes.py
dylan-matthew-turner: I added an error for not being able to find the feed from the branch.

scripts/diff-pkg-changes.py
dylan-matthew-turner: I believe I've caught all the logic errors and either handled them or raised actual errors for them.
Please take another look

CHANGELOG.md
```diff
@@ -39,6 +41,10 @@ Branch: `nilrt/22.5/hardknott`
   - [meta-mingw](https://github.com/ni/nilrt/pull/124).
   - [meta-mono](https://github.com/ni/nilrt/pull/123).
 
+#### Feed Changes
+
+- Changes to [core/](https://github.com/dylan-matthew-turner/nilrt-pkg-changes/blob/nilrt/master/hardknott/docs/feed-changelog.md#changes-to-core)
```
shruthi-ravi: I assume you want to update this link to the ni/nilrt repo? Same below.

scripts/emit-feed-changelog-markdown.py
```diff
@@ -0,0 +1,86 @@
...
+def main(args: Namespace) -> int:
+    # Read the text in as string
+    json_str: str = ''
+    with open(args.input_file) as json_file:
+        json_str = json_file.read()
+
+    # Then split into lines of json
+    json_lines: list[str] = json_str.split('\n')
+    data: dict = {}
+    for line in json_lines:
```
shruthi-ravi: Could you do
`for line in json_file:`? 
That should work as well as reading the file as a string and splitting by newline.

scripts/emit-feed-changelog-markdown.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+    parser.add_argument(
+        'input_file',
+        metavar = 'INPUT', type = str,
+        help = 'The JSON data file to read'
+    )
+    args = parser.parse_args()
+
+    # Run the code now
+    main(args)
```
shruthi-ravi: You probably need to `sys.exit(main(args))` here if you intend to use main()'s return value. Otherwise, do not return anything explicitly from main().

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+from dataclasses import dataclass
+
+# Defines the base URL for feeds
+DEF_FEED_BASE: str = 'http://download.ni.com/ni-linux-rt/feeds'
+
+# Constants used for constructing the url to a file on GitHub containing feed information
+GITHUB_BASE: str = 'https://raw.githubusercontent.com/ni/meta-nilrt'
+GITHUB_CONF_FNAME: str = 'conf/distro/nilrt.conf'
+
+def main(args: Namespace) -> int:
```
shruthi-ravi: Move this function closer to where you call it.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+#     3. Parse the package
+#     4. Compare the packages to each other
+#     5. Generate JSON report
+# - Note: Try to keep as few non-base python packages as possible to keep it portable
+# - Example (finding differences between 8.11 and 9.0):
+#     ./scripts/diff-pkg-changes.py \
+#         nilrt/21.8/sumo 8.11 nilrt/22.5/hardknott 9.0 \
+#         --feed_base http://nickdanger.amer.corp.natinst.com/feeds
+# - Example (updating an existing file):
+#     ./scripts/diff-pkg-changes.py ... >> ./docs/feed-changelog.json
```
shruthi-ravi: I would put this into the usage for the script, as `epilog` to the ArgumentParser logic below.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+# Authors: Dylan Turner
+# Description:
+# - Diff feeds of two NI Linux RT releases from their GitHub branches to create a json diff
+# - Steps:
+#     1. Pull core/ and extra/ feed info from GitHub for branches to compare
+#     2. Pull the Packages files from all four feeds
+#     3. Parse the package
+#     4. Compare the packages to each other
+#     5. Generate JSON report
+# - Note: Try to keep as few non-base python packages as possible to keep it portable
```
shruthi-ravi: PP: Remove this line. I think it is understood.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+        )
+        self.add_argument(
+            'old_branch',
+            metavar = 'OLD', type = str,
+            help = 'The nilrt branch of the older release'
+        )
+        self.add_argument(
+            'old_vers',
+            metavar = 'OLD_ID', type = str,
+            help = 'The version number of the older release'
```
shruthi-ravi: You can get this from the `DISTRO_VERSION` variable in [nilrt.conf](https://github.com/ni/meta-nilrt/blob/777c9c27809d1c5e5c5c735c0d63d5510dfbd1fa/conf/distro/nilrt.conf#L3) as well.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+
+# Defines the base URL for feeds
+DEF_FEED_BASE: str = 'http://download.ni.com/ni-linux-rt/feeds'
+
+# Constants used for constructing the url to a file on GitHub containing feed information
+GITHUB_BASE: str = 'https://raw.githubusercontent.com/ni/meta-nilrt'
+GITHUB_CONF_FNAME: str = 'conf/distro/nilrt.conf'
+
+def main(args: Namespace) -> int:
+    # Get user input
```
shruthi-ravi: I appreciate how thorough you are with helpful comments everywhere, it has helped me understand the flow better. But I think some of them can be deleted without affecting readability. 
For example, this line and the following `Get the feed version`, `Get the packages` are superfluous given that you already have great, self-explanatory function names and variables. See if there are other instances too that you can remove. It's a fine balance, and I'll let you decide on the other instances.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+        self.branch_name = branch_name
+        super().__init__('Failed to find feed \'%s\' in branch %s.' % (feed_name, branch_name))
+
+# Some issue with the Makefile containing the feed name
+class InvalidBranchFileError(PkgDiffError):
+    def __init__(self, branch_name: str, file_name: str) -> 'InvalidBranchFile':
+        self.branch_name = branch_name
+        self.file_name = file_name
+        super().__init__(
+            'Information in file \'%s\' in branch \'%s\' contained unexpected information.'
```
shruthi-ravi: You're missing `branch_name` and `file_name` references here.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+
+# Error handling base class for the logic
+class PkgDiffError(Exception):
+    pass
+
+# When a file within the branch cannot be found
+class BranchFileNotFoundError(PkgDiffError):
+    def __init__(self, branch_name: str, file_name: str) -> 'BranchFileNotFoundError':
+        self.branch_name = branch_name
+        self.file_name = file_name
```
shruthi-ravi: If you aren't using these attributes anywhere else in the class, I suggest removing them. Same for the remaining Exception classes.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+    for line in file.split('\n'):
+        if line.startswith('NILRT_FEED_NAME = '):
+            feed_line = line
+            break
+
+    if len(feed_line) < 21:
+        raise InvalidBranchFileError(branch_name, GITHUB_CONF_FNAME)
+
+    # Extract the info
+    return feed_line[19:-1]
```
shruthi-ravi: Make sure to remove trailing whitespaces from the line.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+    old_extra_pkgs: list['Package'], new_extra_pkgs: list['Package'],
+    old_vers: str, new_vers: str
+) -> dict:
+    # Handle core diffs
+    core_pkg_changes: list[dict] = []
+    core_pkgs_added: list[dict[str, str]] = []
+    core_pkgs_removed: list[dict[str, str]] = []
+    for pkg in old_core_pkgs:
+        other_vers_of_pkg: list['Package'] = \
+            [ i for i, other_pkg in enumerate(new_core_pkgs) if other_pkg.name == pkg.name ]
```
shruthi-ravi: If your `_pkgs` variables were dictionaries instead, with the key being the package name, and the value, the Package object, this operation would be less expensive, right? Right now, for each package in `old_core_pkgs`, you need to iterate over `new_core_pkgs`.
From my memory, our oe-core and oe-extra feeds do not contain packages with several different versions. So the dict should have unique entries. And I think you also indicate that assumption below on line 294.

shruthi-ravi (commented):
Nice work on the scripts! I have only minor suggestions on the scripts.

But I have some rework comments on the commit messages. I should have elaborated on some of them in my previous comment, my bad.
General comments:
- Wrap the commit message text around 70 characters
- Try to limit the commit message header to 50 characters when possible
- Write the commit message and header in the imperative
- All of this are elaborated in a great [guide](https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project), under `Commit Guidelines`

Specific comments:
- Squash the `Remove dead code` commit since it does not provide valuable history
- Some of your commit messages go into detail on how the scripts achieve their ends, which I think is unnecessary given that you already have comments detailing the same in the scripts

Also, please refer to the azdo WI in the PR description.
Thanks!

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+            new_extra_pkgs.pop(other_vers_of_pkg[0])
+    for pkg in new_extra_pkgs:
+        # We removed packages that were duplicates, so what's left is new packages
+        extra_pkgs_added.append({
+            'name': pkg.name,
+            'version': pkg.version
+        })
+
+    return {
+        (old_vers + '->' + new_vers): {
```
amstewart: I don't think we need this top-level mapping, right? The version information is already stored in the `old/new_vers` keys one level down, and I think it would be perfectly valid JSON to use that level as the top-level.

CHANGELOG.md
amstewart: Given that the feeds are going to change slightly between *every* release, I don't see a reason to burden the repo maintainer with adding these lines for every release. I think the line you added at the top of the file directing readers toward the `feed-changelog.md` are sufficient.

scripts/diff-pkg-changes.py
amstewart: He could; but I think I prefer the script as-is. The logic to parse out that variable would be more confusing that it is worth, I think.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+                'old': str(self.size),
+                'new': str(other.size)
+            }
+
+        if diff_dict != {}:
+            diff_dict['name'] = self.name
+
+        return diff_dict
+
+# Get the year/identity of each branch's feed by pulling the info from a file on GitHub
```
amstewart: NIT: Since you're adding annotations for all of your methods here, why not put them in docstrings within the method? That way, python will embed them into the method object's internal documentation structure, instead of just discarding the comment data?
eg.
```python3
def get_feed(branch_name: str) -> str:
    """Get the year/identity of each branch's feed by pulling the info from a file on GitHub"""
    ...
```

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,351 @@
...
+    # Parse every package listing
+    for pkg_txt in file.split('\n\n'):
+        if pkg_txt == '':
+            continue
+
+        pkg_dict: dict[str, str] = {}
+        for line in pkg_txt.split('\n'):
+            if line == '':
+                continue
+            tokens: list[str] = line.split(': ')
```
amstewart: I don't think there is anything which keeps opkg index field bodies from containing instances of the substring `: `. So [this logic](https://docs.python.org/3/library/stdtypes.html#str.split) might return >2 tokens, and then `tokens[2:]` would get thrown away by a few lines later.

Could you do this instead?
```python3
tokens: list[str] = line.split(': ', maxsplit=1)
```

scripts/diff-pkg-changes.py
amstewart: I don't think its safe to say that the feeds couldn't contain two packages with the same name, but different versions. In fact, you could reasonably have: same name, same version, different architecture. So when considering whether a package matches, it should only match if the architectures are the same.

scripts/diff-pkg-changes.py
amstewart: I think its likely that we will want to change the contents of this data structure at some point in the future. Can we add a "version" field, which expresses the version of the data structure?

amstewart: This structure also needs to encode the architecture of the feed. We only support `x64` in hardknott, for now. But that might well change in the future, and we'll need to be able to distinguish between architectures.

scripts/emit-feed-changelog-markdown.py
```diff
@@ -0,0 +1,86 @@
...
+        + '## Description\n\n' \
+        + 'This documents the full changes to the core/ and extra/ feeds for each version ' \
+        + 'release.\n\n' \
+        + 'The changes are too long to keep in the [Changelog](/CHANGELOG.md) itself, ' \
+        + 'so they are kept here\n\n' \
+        + 'Generate them by:\n    1. adding a line to ./docs/feed-changelog.json: ' \
+        + '`./scripts/diff-pkg-changes.py <branch of old version> <old version> >>' \
+        + '<branch of new version> <new version> >> ./docs/feed-changelog.json`\n    ' \
+        + '2. updating the markdown with `./scripts/emit-feed-changelog-markdown.py ' \
+        + 'feed-changelog.json > feed-changelog.md`\n\n'
```
amstewart: You can use triple-quote notation here to retain whitespace and avoid having to do the escape shenanigans.
```python3
FEED_CHANGELOG_START: str = """# Changes to NI Linux RT Package Feeds

## Description
This documents the full changes to the core/ and extra/ feeds for each version
...
"""
```

scripts/emit-feed-changelog-markdown.py
```diff
@@ -0,0 +1,86 @@
...
+    # Create a report from that info
+    report: str = FEED_CHANGELOG_START
+    for change_name, change_data in data.items():
+        report += \
+            '## Feed Changes from Version ' \
+                + change_data['old_vers'] + ' to Version ' + change_data['new_vers'] \
+                + '\n\n### Changes to core/\n\n* Added:\n'
+        for pkg in change_data['core']['added']:
+            if not pkg['name'].endswith('-lic'):
+                report += '    * ' + pkg['name'] + '\n'
```
amstewart: You can define a submethod within `main()` like
```python3
def main(...):
    MASKED_SUFFIXES = ["-lic"]
    ...
    def report_changes(feed_name, operation):
        report += f"* {operation}\n"
        for pkg in change_data[feed_name][operation]:
            if not any([pkg['name'].endswith(suffix) for suffix in MASKED_SUFFIXES]):
                report += f"\t* {pkg["name"]}\n"

    ...
        report += \
            '## Feed Changes from Version ' \
                + change_data['old_vers'] + ' to Version ' + change_data['new_vers'] \
                + '\n\n### Changes to core/\n\n'
        report_changes('core', 'added')
        report_changes('core', 'removed')
        ....
```

scripts/emit-feed-changelog-markdown.py
```diff
@@ -0,0 +1,86 @@
...
+    # Add an argument for the JSON file
+    parser = ArgumentParser(
+        description = 'Convert JSON changelog data into a readable report'
+    )
+    parser.add_argument(
+        'input_file',
+        metavar = 'INPUT', type = str,
+        help = 'The JSON data file to read'
+    )
+    args = parser.parse_args()
```
amstewart: Just read the json list from stdin. Stdin is your friend. :)

```python3
for line in sys.stdin:
    json_line = json.loads(line)
...
```

Then you don't need the argparsing bits at all.

amstewart (changes requested):
I like the new changelog much better. :+1: 

+1 most of Shruthi's commentary as well.

scripts/diff-pkg-changes.py
dylan-matthew-turner: Yeah you're right we don't need this. It's is from before I knew to store as individual lines.

scripts/diff-pkg-changes.py
dylan-matthew-turner: Perhaps the key could be constructed by concatenating name + architecture and using that as a dictionary to be less expensive.
Something like new_core_pkgs.get(pkg.name + '\_' + pkg.architecture, None), and in the get_packages, have it do something like dict[pkg.name + '\_' + pkg.architecture] = pkg

dylan-matthew-turner: __Example:__

In `get_packages`:
```
pkgs[pkg_dict['Package'] + '_' + pkg_dict['Architecture']] = Package(
    pkg_dict['Package'],
    pkg_dict['Version'],
    pkg_dict.get('Depends', ''),
    pkg_dict.get('Provides', ''),
    ...
)
```

In `gen_report`:
```
for pkg_id, pkg in old_core_pkgs.items():
    other_vers_of_pkg = new_core_pkgs.get(pkg_id, None)
    if other_vers_of_pkg is None:
        core_pkgs_removed.append({
            'name': pkg.name,
            'version': pkg.version
        })
    else:
        other_pkg: 'Package' = new_core_pkgs[pkg_id]
        diff: dict = pkg.diff(other_pkg)
        if diff != {}:
            core_pkg_changes.append(diff)
        new_core_pkgs.pop(pkg_id)
for pkg_id, pkg in new_core_pkgs.items():
    core_pkgs_added.append({
        'name': pkg.name,
        'version': pkg.version
    })
```

scripts/diff-pkg-changes.py
shruthi-ravi: I noticed this too and wondered if we wanted this for sumo, but I think for a first pass on this script we can make an assumption that we care only about x64. i.e. as long as the architecture doesn't need a big rewrite. 
If we don't care about sumo, for the foreseeable future, there's only x64 and I don't think we should spend too much time accommodating something else. @amstewart ?

dylan-matthew-turner: There's only one place affected by architecture afaik, and that's part of the feed URL used for getting the Packages file. It has a '/x64/' right now. I think that's the only thing that would need to change to adjust architecture

shruthi-ravi: I didn't think of the different 'tunes' - `core2-64`, `x64` and `all`. I don't remember how that is handled by this script, if they all end up in the same list or not.

scripts/diff-pkg-changes.py
shruthi-ravi: I'm not sure why it would be confusing. I think it is more work for the user to look up those versions and pass it to the script. Let's talk offline to see what I'm missing.

dylan-matthew-turner: I think it would just be
```
if line.startswith('DISTRO_VERSION = '):
    distro_line = line
    break
```

scripts/diff-pkg-changes.py
shruthi-ravi: @dylan-matthew-turner is it correct that you're parsing the Packages file only under http://nickdanger.amer.corp.natinst.com/feeds/2022Q3/x64/main/all/
and not also
http://nickdanger.amer.corp.natinst.com/feeds/2022Q3/x64/main/core2-64/ , and
http://nickdanger.amer.corp.natinst.com/feeds/2022Q3/x64/main/x64/ ?

If that's the case, you need to parse them all.

scripts/diff-pkg-changes.py
shruthi-ravi: Looks good to me, except in the `else`, you don't need a second access of the dict. You could just do 
`diff: dict = pkg.diff(other_vers_of_pkg)`.

scripts/diff-pkg-changes.py
amstewart: > I think for a first pass on this script we can make an assumption that we care only about x64. i.e. as long as the architecture doesn't need a big rewrite.

I think it should only take one line or so to include the arch in the JSON structure. I think that's worth the effort at this stage, so that we don't have to rev the JSON structure version in the future.

CHANGELOG.md
dylan-matthew-turner: It has been removed altogether as this is unnecessary work for future maintainers

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,353 @@
...
+DEF_FEED_BASE: str = 'http://download.ni.com/ni-linux-rt/feeds'
+
+# Constants used for constructing the url to a file on GitHub containing feed information
+GITHUB_BASE: str = 'https://raw.githubusercontent.com/ni/meta-nilrt'
+GITHUB_CONF_FNAME: str = 'conf/distro/nilrt.conf'
+ARCH: str = 'x64'
+SUB_FEEDS: list[str] = [ 'x64', 'core2-64', 'all' ]
+
+# If you change the info in the report, update this variable
+DATA_STRUCT_VERS = '0'
```
shruthi-ravi, line 29: Can you elaborate on this? I didn't follow it's usage.

scripts/emit-feed-changelog-md.py
```diff
@@ -0,0 +1,77 @@
...
+            for pkg in data[feed_name][operation]:
+                if not any([ pkg['name'].endswith(suffix) for suffix in MASKED_SUFFIXES ]):
+                    local_report += f'    - {pkg["name"]}\n'
+            return local_report
+
+        report += \
+            '## Feed Changes from Version ' \
+                + data['old_nilrt_version'] + ' to Version ' + data['new_nilrt_version'] \
+                + '\n\n### Changes to core/\n\n'
+
```
shruthi-ravi, line 54: Where are the `changes` in packages?

shruthi-ravi (approved):
Everything looks good! Minor comments and I missed mentioning that each of your commit messages need to contain your "signoff" at the end:
Signed-off-by: Name \<email ID\>

scripts/diff-pkg-changes.py
dylan-matthew-turner: Because the data structure of the final report can change over time, we want to keep track of what "version" of the data structure we're currently using. If someone adds a field to the report, they should increase this value. It gets added to the report as `"version": DATA_STRUCT_VERS`, and since this is the initial version, so I put zero.

scripts/emit-feed-changelog-md.py
```diff
@@ -0,0 +1,77 @@
...
+# Authors: Dylan Turner
+# Description:
+# - Convert JSON changelog data into a human-readable report
+# - Example:
+#     ./emit-feed-changelog-md.py <./docs/feed-changelog.json > ./docs/feed-changelog.md
+# - Note: Currently we only care about packages added/removed, but more info can be extracted from
+#   the JSON
+
+import sys
+from argparse import ArgumentParser, Namespace
```
amstewart: Unused import?

scripts/emit-feed-changelog-md.py
dylan-matthew-turner: There's another comment talking about this, but essentially, for the "cleaner" report (the markdown) we want to keep it simple and just show what packages have been added/removed

That's part of why the diff-pkg-changes.py script outputs JSON instead. The feed-changelog.json has all changes to the feed in a parseable format, whereas the feed-changelog.md has a short report of just what users will want to see which right now, is just added/removed.

scripts/diff-pkg-changes.py
```diff
@@ -0,0 +1,353 @@
...
+
+class CLIParser(ArgumentParser):
+    '''
+    This is a class that inherits from the ArgumentParser to separate the CLI argument logic
+    from the main function for more readable code.
+
+    The arguments passed in include the old and new branches to compare along with their version
+    numbers. Optionally, one can change the base feed URL
+
+        '''
```
amstewart: NIT: Whitespace.

docs/feed-changelog.md
```diff
@@ -0,0 +1,12921 @@
...
+
+Generate them by:
+    1. adding a line to ./docs/feed-changelog.json: `./scripts/diff-pkg-changes.py <branch of old version> <branch of new version> >> ./docs/feed-changelog.json`
+    2. updating the markdown with: `./scripts/emit-feed-changelog-md.py < ./docs/feed-changelog.json > ./docs/feed-changelog.md`
+
+## Feed Changes from Version 8.11 to Version 9.0
+
+### Changes to core/
+
+* added:
```
amstewart, line 17: I don't think this output is correct. :/ I see (a) way too many packages and (b) many packages listed here which should have been in both the 8.11 and 9.0 feed.

amstewart, line 17: Could you sort the packages by name before adding them to the report? It will help users to find the package they're looking for.

amstewart (changes requested):
Commits look good now. Just remember to add your sign-off. :)

docs/feed-changelog.md
dylan-matthew-turner: I agree it seems like a lot. I'm looking into it, but I haven't had any luck finding an issue. Can you give an example of a package listed that should be in both but is showing as not?

I figured the increase from earlier code versions was from switching from just using http://download.ni.com/ni-linux-rt/feeds/2021.8/x64/main/all to also including /main/x64 and /main/core2-64, but it definitely could be an issue instead.

dylan-matthew-turner: There's also kernel packages clogging stuff up, since even their package name is versioned, e.g. stuff like kernel-module-whatever-5.x.x, but it's not unreasonable to mark those changes.

dylan-matthew-turner: One thing that seems to affect it is that some packages are marked as "removed" when they change architecture. That doesn't seem to happen with "added," but it could

amstewart: Yeah; after a more careful review: this list does actually seem accurate. A lot of packages genuinely got added or removed between 8.X (sumo) and 9.0 (hardknott). And many more changed from the extra/ feed to core/ - which causes them to show up under both added and removed.

So I guess this is fine.

amstewart (approved):

PR #167, opened 2022-08-04
upstream_merge:Add script to merge latest upstream

chaitu236:
Add a script upstream_merge.sh that can be used to automate upstream
merges.
Also add a conf file for nilrt repos.

### Testing
Ran script on OE sources

image

image

On a test repo with merge conflicts
image

On a test repo with no merge conflicts
image


scripts/dev/upstream_merge/upstream_merge.sh
```diff
@@ -0,0 +1,206 @@
...
+            usage
+            ;;
+      esac
+   done
+   shift $((OPTIND-1))
+}
+
+sanity_test_repo() {
+   LOCAL_BASE_BRANCH=$1
+   if ! `git rev-parse --verify $LOCAL_BASE_BRANCH &> /dev/null`; then
```
amstewart: NIT: Could you prefer to use the new subshell syntax `$(...)` rather than the old tilde syntax; it's generally cleaner.

scripts/dev/upstream_merge/upstream_merge.sh
```diff
@@ -0,0 +1,206 @@
...
+DIR=$(dirname ${BASH_SOURCE})
+CONF_FILE="$DIR/repos.conf"
+
+REMOTE_REPO_NAME="automerge_upstream"
+LOCAL_UPSTREAM_BRANCH_PREFIX="dev/automerge/upstream"
+LOCAL_BRANCH_NAME="dev/automerge/ni"
+
+usage() {
+   echo "Usage: $0 [-c <conf file>][-h] " 1>&2
+   exit 1
```
amstewart: MINOR: `upstream_merge.sh -h` should return `0`, rather than `1`. You can just move the `exit` call here into the `parse_args()` case statement.

scripts/dev/upstream_merge/upstream_merge.sh
```diff
@@ -0,0 +1,206 @@
...
+main() {
+
+   while read -u 10 line; do
+      if [[ "$line" =~ ^#.* ]]; then
+         continue
+      fi
+      LOCAL_REPO=`echo $line | awk '{print $1}'`
+      UPSTREAM_REPO=`echo $line | awk '{print $2}'`
+      UPSTREAM_BRANCH=`echo $line | awk '{print $3}'`
+      UPSTREAM_BRANCH=`echo $line | awk '{print $3}'`
```
amstewart: Duplicate line?

scripts/dev/upstream_merge/upstream_merge.sh
```diff
@@ -0,0 +1,206 @@
...
+      echo " ... ERRORS"
+   fi
+}
+
+handle_repo() {
+   LOCAL_REPO=$1
+   UPSTREAM_REPO=$2
+   UPSTREAM_BRANCH=$3
+   LOCAL_BASE_BRANCH=$4
+   LOCAL_UPSTREAM_BRANCH="$LOCAL_UPSTREAM_BRANCH_PREFIX/$UPSTREAM_BRANCH"
```
amstewart: You can use the `local <var_name>=<value>` syntax to create function-scoped variables, so that they don't pollute the global scope.

scripts/dev/upstream_merge/upstream_merge.sh
```diff
@@ -0,0 +1,206 @@
...
+            ;;
+         *)
+            usage
+            ;;
+      esac
+   done
+   shift $((OPTIND-1))
+}
+
+sanity_test_repo() {
```
amstewart, line 48: Could you alphabetize the functions in this script so that they're easier to traverse?

scripts/dev/upstream_merge/upstream_merge.sh
amstewart: You can probably replace the `/dev/null` redirection with the `git rev-parse --quiet` option.

scripts/dev/upstream_merge/upstream_merge.sh
```diff
@@ -0,0 +1,206 @@
...
+      exit 1
+   fi
+}
+
+add_remote() {
+   UPSTREAM_REPO=$1
+
+   git remote remove $REMOTE_REPO_NAME &> /dev/null || true
+   git remote add $REMOTE_REPO_NAME $UPSTREAM_REPO
+}
```
amstewart: I'm uneasy about this script modifying a submodule's remotes as a consequence of its action. I think we should honor however the user has their remotes configured, even if that means that the user needs to specify what they have named their upstream and NI GH remotes.

scripts/dev/upstream_merge/upstream_merge.sh
```diff
@@ -0,0 +1,206 @@
...
+
+sync_remote_branch() {
+   UPSTREAM_BRANCH=$1
+   LOCAL_UPSTREAM_BRANCH=$2
+
+   git branch -D $LOCAL_UPSTREAM_BRANCH &> /dev/null || true
+   if ! `git fetch $REMOTE_REPO_NAME $UPSTREAM_BRANCH:$LOCAL_UPSTREAM_BRANCH &> /dev/null`; then
```
amstewart: Is any of this branch manipulation actually necessary to create a rebased branch? I don't think it is.

amstewart (changes requested):
I think we're duplicating some of what git should be doing for us wrt. remote management. I think the input data in `repos.conf` can be simplified to:
```
# submodule_path    upstream_ref    base_ref
sources/bitbake     upstream/1.50   gh-ni/nilrt/master/hardknott
```

ie. The remote name can be prefixed to the refs and we can rely on the user to have setup their remotes correctly. If we feel like that is too burdensome, I would support having a separate script which sets up the submodule remotes in some standardized way, rather than rolling that management into this script.

amstewart:
Looking at the docs for `git submodule update --merge`, it seems like a lot of what you're doing here might already be built into the git tooling. Have you considered using that command, and some of the `git-submodule` options that allow you to control how the update occurs (and maybe a composed .gitmodules file) to accomplish this same thing?

scripts/dev/upstream_merge/upstream_merge.sh
chaitu236: My intent with this script was that however the user have configured their repos, the script would setup whatever it needs in its own namespace (`automerge` in this case. i.e., remotes are `automerge_upstream`, and branches are `dev/automerge/..`) and then do the merge.

On a newly setup nilrt repo, the user doesn't have upstream remotes setup and I don't think it should be a requirement for them to do that to use this script as that can be automated as well.

> I'm uneasy about this script modifying a submodule's remotes as a consequence of its action

The script only adds a new remote and shouldn't affect what the developer has configured for their use. The remote setup here is intended to be used only in the script (hence the namespace).

If your concern is about polluting remote list, we can consider adding a `--cleanup` option to the script that would go ahead and remove the remote (and `dev/automerge/...` branches too?).

scripts/dev/upstream_merge/upstream_merge.sh
chaitu236: Are you suggesting we don't create `LOCAL_UPSTREAM_BRANCH` but just fetch `REMOTE_REPO_NAME/UPSTREAM_BRANCH` and use it directly?

chaitu236:
> Looking at the docs for `git submodule update --merge`, it seems like a lot of what you're doing here might already be built into the git tooling. Have you considered using that command, and some of the `git-submodule` options that allow you to control how the update occurs (and maybe a composed .gitmodules file) to accomplish this same thing?

I wanted this script to be generic enough to be used for a single repo that isn't a submodule of another repo. So that we could extend it in future to work for linux.git (which currently doesn't work because the script doesn't support merging from tags).

scripts/dev/upstream_merge/upstream_merge.sh
chaitu236: All the functions (except main) are currently laid out in the order in which they're called. I find that this is simpler to navigate as we can just read the file top down to understand code flow without digging inside the top level functions.

scripts/dev/upstream_merge/upstream_merge.sh
chaitu236: Apparently `--quiet` is only to quiet error message; it still outputs the hash on success so we still need `/dev/null`.

scripts/dev/upstream_merge/upstream_merge.sh
shruthi-ravi: I would also like the setting up of remotes for all the OE layers to be automated, and not something I should have to do separately.

scripts/dev/upstream_merge/upstream_merge.sh
chaitu236: Got rid of `LOCAL_UPSTREAM_BRANCH` as discussed

scripts/dev/upstream_merge/upstream_merge.sh
chaitu236: Done

scripts/dev/upstream_merge/upstream_merge.sh
chaitu236: Done

scripts/dev/upstream_merge/upstream_merge.sh
chaitu236: Done

scripts/dev/upstream_merge/upstream_merge.sh
chaitu236: Done

amstewart (approved):

shruthi-ravi (approved):

PR #178, opened 2022-09-15
Create scripts to estimate sizes of nilrt images

mike-petersen-ni:
These Bash scripts allow developers to predict approximate disk usage for runmode and safemode images.

More information can be found in the added README.

Output of `./estimate-usage.sh` (tab stops at 12 columns for clarity):

    x86_64      runmode     disk footprint          911M
    x86_64      safemode    disk footprint          40M
    x86_64      safemode    kernel + ramdisk        173M
    ARMv7-A     runmode     disk footprint          151M
    ARMv7-A     safemode    disk footprint          26M
    ARMv7-A     safemode    kernel + ramdisk        108M

On a PXIe-8880, after formatting the disk via MAX, extracting the x86_64 runmode image via SSH results in 899.4M of disk space used.
On a cRIO-9067, after formatting the disk via MAX, extracting the ARMv7-A runmode image via SSH results in 146M of disk space used.

I believe the discrepancy arises from the contents of `/boot`, which the estimations script do not exclude but real targets extract to a different filesystem. The size of that directory is roughtly corresponding to the discrepancy (~10M and ~5M for x86_64 and ARMv7-A).

[AB#1978832](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1978832)

### DIFF
--- a/scripts/dev/estimate-usage/README.md
+++ b/scripts/dev/estimate-usage/README.md
@@ -1,0 +1,103 @@
+# `estimate-usage` scripts
+These scripts exist to allow developers to determine approximately how much disk
+space will be used by a system image when it is deployed.
+
+Note that the sizes returned are **approximate**. The estimation scripts don't
+necessarily use the same software versions or command-line invocations that
+happen on real targets.
+
+## The platform-specific scripts
+The `estimate-usage.<platform>.<image-type>.sh` scripts in this directory are
+designed to consume system images and determine the disk space usage if their
+contents are unpacked into the default filesystem of the platform.
+
+### Running
+These scripts consider every argument given to be an image to provide an
+estimate for. If no arguments are given, they read from standard input. There is
+no argument parsing: there are no command-line options, and an argument of `-`
+does not indicate standard input.
+
+### Privileges
+These scripts assume that they will be executed with sufficient privileges to
+perform mounts, write to device files, and (un)load kernel modules. For this
+reason, they are intended to be run within a virtual machine. They will attempt
+to clean up after themselves, but **may have unintended consequences**: for
+instance, the runmode estimation script for ARMv7-A assumes that it creates the
+ubi flash device `/dev/ubi0` and writes to it, so if that device already exists
+you may suffer data loss.
+
+### Software requirements
+Running the scripts for ARMv7-A requires MTD and U-Boot tools for runmode and
+safemode, respectively. On Debian, the packages can be installed with:
+
+    apt-get install mtd-utils u-boot-tools
+
+### Output format
+The output of these scripts is a five-column TSV (tab-separated value) table on
+standard output. The columns have the following meanings:
+
+ 1. The name of the image examined
+ 2. The CPU architecture (`x86_64` or `ARMv7-A`, assumed by the script)
+ 3. The type of image (`runmode` or `safemode`, assumed by the script)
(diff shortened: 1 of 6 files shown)


shruthi-ravi:
@mike-petersen-ni  can you squash the commits where logically possible? They would not be very useful in the overall nilrt repo history as they are now. Can you also add the output of the script and compare it to the actual sizes on targets in the PR description?

scripts/dev/estimate-usage/estimate-usage.x86_64.runmode.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+set -e
+
+print_fs_usage() {
+    local image="$1"
+    local fs="$(mktemp)"
+    truncate -s 1G "$fs"
```
shruthi-ravi: I think you could increase this to 2G since the size is already so close to 1G. Is there an x64 target with only 1G disk space?

scripts/dev/estimate-usage/estimate-usage.armv7-a.safemode.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+set -e
+
+print_fs_usage() {
+    local input_image="$1"
+    local image="$input_image"
+    if test ! -f "$image"; then
+	# We need to look at its filesize and read it multiple times, so if it's
```
shruthi-ravi: I don't follow this. Can you explain?

scripts/dev/estimate-usage/README.md
```diff
@@ -0,0 +1,103 @@
+# `estimate-usage` scripts
+These scripts exist to allow developers to determine approximately how much disk
```
shruthi-ravi, line 2: Minor: Update commit header to `Add README for the estimate-usage scripts`. Try to keep the commit header length under 50 for all commits.

shruthi-ravi (approved):
Good work! Thanks for your help.

scripts/dev/estimate-usage/estimate-usage.x86_64.runmode.sh
mike-petersen-ni: Makes sense to me. Will be in next push.

scripts/dev/estimate-usage/estimate-usage.armv7-a.safemode.sh
mike-petersen-ni: These scripts can be used as part of a pipeline (in fact the top-level script uses them that way to save some disk I/O for the VM). Unlike the others, this subscript involves reading the file size (with `du --apparent-size` a couple lines later) and its content twice (with `dumpimage`). If the image given is not a plain file (e.g. a symbolic link or FIFO) the script will fail (producing an incorrect answer for the file size, and for FIFOs it will fail with an invalid format error for the second invocation of `dumpimage`, as the first invocation consumed the input already).  
It's probably possible to use `tee` to avoid this, but I figured that'd be more difficult to read.

The other scripts are unaffected by this consideration as they only read the image once, when unpacking the tarball.

scripts/dev/estimate-usage/README.md
mike-petersen-ni: Shortened other commit messages too.

amstewart (approved):

PR #182, opened 2022-09-20
[hardknott] Update feed changelogs for 22.8 release branch.

SparkingSpork:
Updating feed changelogs for 22.5 to 22.8 changes.

Signed-off-by: Charlie Johnston <charlie.johnston@ni.com>

AzDO Work Item: https://dev.azure.com/ni/DevCentral/_workitems/edit/2006459/

## Testing:
N/A. Here are the commands I ran:
```
./scripts/diff-pkg-changes.py nilrt/22.5/hardknott nilrt/22.8/hardknott --feed_base http://nickdanger.amer.corp.natinst.com/feeds >> ./docs/feed-changelog.json  
./scripts/emit-feed-changelog-md.py <./docs/feed-changelog.json > ./docs/feed-changelog.md
```


SparkingSpork:
**Note:** Doing this now just so it doesn't block us later. This will need performed again before Friday if we're able to get the persistent logs changes cherry-picked as that adds a new package as well.

amstewart (approved):
We should really filter out the kernel module packages. They're like 90% of the diff and they're going to change every release, just because the kernel ABI will change slightly.

shruthi-ravi (approved):

PR #185, opened 2022-09-28
[sumo] Updated README and container scripts for external builders.

SparkingSpork:
Currently the instructions for building sumo don't work for external builds due to changes and the addition of pyrex to the sumo branch. These fixes bring the README and container creation scripts more in-line with expected behavior.
- Updated the `create-build-nilrt.sh` script to tag the container with the branch codename instead of `latest`.
- Updated the README to include steps necessary for pyrex and a comment on potential upstream branch name changes.

## Testing:
Built the core and extra package groups locally without the `--org` flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,18 @@ Build Steps
         git submodule init
         git submodule update --remote --checkout
 
-2. Set up your shell environment to run bitbake:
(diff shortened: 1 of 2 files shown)


amstewart (approved):
I'll cherry pick this into the academic mainline as well, once it is approved.

shruthi-ravi (approved):

PR #186, opened 2022-09-29
Add script to diff CVE .csv files

shruthi-ravi:
We would like to track the CVEs added in a new release, in addition to tracking all the CVEs every release. 
Add a script that generates the diff of two csv files and writes it to a csv file.

AzDO [2131135](https://ni.visualstudio.com/DevCentral/_workitems/edit/2131135)

### DIFF
--- a/scripts/dev/diff-cve-csvs.py
+++ b/scripts/dev/diff-cve-csvs.py
@@ -1,0 +1,77 @@
+#!/usr/bin/python3
+
+import argparse
+import csv
+from pprint import pprint
+
+def get_cves(filename):
+    with open(filename, newline='') as fp:
+        dialect = None
+        try:
+            dialect = csv.Sniffer().sniff(fp.read(1024), delimiters=",\t;")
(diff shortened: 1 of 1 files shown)


scripts/dev/diff-cve-csvs.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+#print("new_cves_added")
+#pprint(new_cves_added)
+#print("existing_cves_added_to_new_pkg")
+#pprint(existing_cves_added_to_new_pkg)
+#print("removed_cves")
+#pprint(removed_cves)
+
+
```
amstewart, line 77: Do you want to remove this block before we merge?

shruthi-ravi: I wanted to keep it so that it is explicit to whoever was doing this next time that the script is extracting (but not printing) this additional information that they could look at.

amstewart (approved):

PR #197, opened 2022-11-02
scripts/pipelines: Create script for building meta-toolchain

mike-petersen-ni:
This creates a simple script to easily allow building of the cross-compile toolchain, similar to the other scripts used for pipelines.

Note: running the build may be dependent on ni/meta-nilrt#494.

[AB#1242721](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/1242721)

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,15 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
 
     These scripts are also a good source for understanding the steps to build a package feed manually. Note that if you are building package feeds *manually*, you must bitbake the special `package-index` target before using the feed.
 
+    * ##### Building the cross-compile toolchain
+        One of these pipelines can build the cross-compile toolchain. By default, it builds for x86_64 Linux hosts.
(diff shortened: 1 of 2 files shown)


scripts/pipelines/build.toolchain-feeds.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
shruthi-ravi: Should the name of the script be just `build.toolchain`? I don't think we need the `feeds`.

shruthi-ravi (changes requested):

scripts/pipelines/build.toolchain-feeds.sh
mike-petersen-ni: I'm fine with that. Will be in the next push.

shruthi-ravi (approved):

amstewart (approved):

PR #199, opened 2022-11-10
scripts: Add script to build toolchain for Windows

shruthi-ravi:
Built successfully.

AzDO [2105299](https://ni.visualstudio.com/DevCentral/_workitems/edit/2105299)


scripts/pipelines/build.cross-toolchain.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+SCRIPT_ROOT=$(realpath $(dirname $BASH_SOURCE))
+
+. "${SCRIPT_ROOT}/build.common.sh" "$@"
+
+echo "INFO: Setting SDKMACHINE to x86_64-mingw32"
+sed -i '/^SDKMACHINE/{s/\"x86_64\"/\"x86_64-mingw32\"/;}' ./conf/local.conf
```
mike-petersen-ni: Nit: technically don't need the block or the escapes in the sed script.

```suggestion
sed -i '/^SDKMACHINE/s/"x86_64"/"x86_64-mingw32"/' ./conf/local.conf
```

shruthi-ravi: Done in v2.

shruthi-ravi:
Update v2:

Addressed Mike's comment on blocks and escapes in sed.

scripts/pipelines/build.cross-toolchain.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+SCRIPT_ROOT=$(realpath $(dirname $BASH_SOURCE))
+
+. "${SCRIPT_ROOT}/build.common.sh" "$@"
+
+echo "INFO: Setting SDKMACHINE to x86_64-mingw32"
+sed -i '/^SDKMACHINE/s/"x86_64"/"x86_64-mingw32"/' ./conf/local.conf
```
amstewart: Instead of using `sed` - which is relatively unreliable - can we instead add `SDKMACHINE` to `BB_ENV_EXTRAWHITE`/`BB_ENV_PASSTHROUGH_ADDITIONS` and then just set its value in the calling script, in that same way that you would set the `MACHINE`?

amstewart (changes requested):

shruthi-ravi:
Update v3:

SDKMACHINE is now set with the BB_ENV_EXTRAWHITE mechanism.
@amstewart

amstewart (approved):

PR #208, opened 2023-01-03
repos: re-add meta-mingw to upstream sources list

dylan-matthew-turner:
It shouldn't be missing as the submodule is there and we keep track of the changes.


dylan-matthew-turner:
Actually, this is not quite correct. I will correct in just a minute

dylan-matthew-turner:
The local branch for meta-mingw appears to be wrong according to the upstream merge script. I'll correct it

PR #210, opened 2023-01-04
pyrex.ini: Set '--security-opt seccomp=unconfined'

chaitu236:
This is needed for building kirkstone feeds on build agents with old libseccomp.

Azdo WI: [2105148](https://dev.azure.com/ni/DevCentral/_workitems/edit/2105148)

### Testing
PR build pipeline builds OE with these changes.

### DIFF
--- a/scripts/azdo/conf/pyrex.ini
+++ b/scripts/azdo/conf/pyrex.ini
@@ -101,6 +101,8 @@ envvars =
 # Extra arguments to pass when running the image. For example:
 #   --mount type=bind,src=${env:HOME}/.ssh,dst=${env:HOME}/.ssh,readonly
 #   --device /dev/kvm
+args =
+	--security-opt seccomp=unconfined
 
 # Prefix for all Pyrex commands. Useful for debugging. For example:
 #   strace -ff -ttt -o strace.log --


gratian (approved):

pyrex.ini
```diff
@@ -103,6 +103,7 @@ envvars =
 #   --device /dev/kvm
 args =
 	--device /dev/kvm
```
chaitu236: > --device /dev/kvm

@amstewart Why do we need this?

amstewart: The [build.vms.sh](https://github.com/ni/nilrt/blob/nilrt/master/hardknott/scripts/pipelines/build.vms.sh) pipeline script is [run within the pyrex container](https://github.com/ni/nilrt/blob/nilrt/master/hardknott/scripts/pipelines/build.vms.sh#L207) and uses QEMU/KVM to create the NILRT VM artifacts. The `--device /dev/kvm` arg just enables the pyrex container to use the container-host's KVM device, if present.

amstewart (approved):

pyrex.ini
chaitu236: Ok, so why didn't we add it to azdo pyrex.ini also?

chaitu236: Is kvm device not available on azdo agents?

amstewart: It might not have been. I think I attempted to add it to both when I originally authored this stuff, but that was over a year ago. The host configs may have changed.

pyrex.ini
```diff
@@ -103,6 +103,7 @@ envvars =
...
+	--security-opt seccomp=unconfined
```
amstewart: Actually, I don't like enabling this option in the developer `pyrex.ini`. I think it makes better security sense to have host machines which encounter the same errors as @gratian's machine, add this option to their local INI and not check this into the mainline for everyone.

The `.../azdo/.../pyrex.ini` file I'm less worried about, since that should only be used on our machines.

amstewart (changes requested):

pyrex.ini
chaitu236: Seems like its available now. I'll make that a separate change once I test it.

pyrex.ini
chaitu236: Ok, removed `pyrex.ini` changes.

amstewart (approved):

PR #211, opened 2023-01-05
upstream_merge: new argument enabling submodule auto-checkouts

dylan-matthew-turner:
When performing an upstream merge on a non-default release, one must perform a git checkout on each submodule as the local branches are not auto-grabbed; they are simply set to a remote commit.

This adds optional functionality to upstream_merge.sh which forces a git checkout on the provided local branch from repos.conf before proceeding with a sanity check and the rest of the operations.

Default behavior of not doing this checkout is preserved and helpful messages have been added when this new behavior is being used. The messages look like this in practice:

```
sources/bitbake
    Non-default checkout enabled. Attempting to do `git checkout nilrt/master/kirkstone`.
    Checkout succeeded.

    Branch dev/automerge/ni already exists
    Delete Branch(d)/Skip Repo(s)/Cancel Merge(c)?
```

### DIFF
--- a/scripts/dev/upstream_merge/upstream_merge.sh
+++ b/scripts/dev/upstream_merge/upstream_merge.sh
@@ -4,7 +4,11 @@
 # repos described in repos.conf
 #
 # For each local repo, it does following
-# 1. Check if local_base_branch specified in repos.conf exists
+# 1. [Skippable] Check if local_base_branch specified in repos.conf exists
+#    a. When trying to merge with upstream on a branch that's not the default, one first needs to
+#    run checkout in each submodule. We already run a checkout but typically do a sanity check
+#    first. In the case you aren't on the default release, you can add -f to go ahead and force the
+#    checkout, skipping the sanity check
 # 2. Checkout local_base_branch and run git pull
 # 3. Add a 'remote' named 'automerge_upstream'
 #    a. If the remote already exists, remove it first
@@ -19,6 +23,7 @@ CONF_FILE="$DIR/repos.conf"
 
 REMOTE_REPO_NAME="automerge_upstream"
 LOCAL_BRANCH_NAME="dev/automerge/ni"
+FORCE_CHECKOUT=""
 
 usage() {
    echo "Usage: $0 [-c <conf file>][-h] " 1>&2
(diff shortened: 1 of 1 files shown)


amstewart (approved):

scripts/dev/upstream_merge/upstream_merge.sh
```diff
@@ -48,6 +56,18 @@ parse_args() {
 sanity_test_repo() {
    local LOCAL_BASE_BRANCH=$1
 
+   # Optionally, attempt to checkout a branch before the sanity check (in case not using default)
```
chaitu236: Am I correct in understanding that this change is equivalent to just skipping `sanity_test_repo` step since the next step `update_local_base_branch` already does `git checkout $LOCAL_BASE_BRANCH `?

And if that's the case, why not just have this option skip the `sanity_test_repo` step? I think `git rev-parse --verify $LOCAL_BASE_BRANCH` ceases to be useful anymore after we do a checkout.

dylan-matthew-turner: I agree. That does make more sense. I'll adjust it

scripts/dev/upstream_merge/upstream_merge.sh
```diff
@@ -13,25 +13,34 @@
 #    a. If the branch name exists, ask to delete or skip or cancel
 # 6. Merges 'automerge_upstream/$LOCAL_BRANCH_NAME' into 'dev/automerge/ni'
 #    a. Success or failure is reported; script continues in either case
+#
```
chaitu236: Can you move this to under step 1 above - something like

1. [Skippable] Check if local_base_branch specified in repos.conf exists
    - When trying to merge with upstream on a new branch that's not the default, one needs to run a
       checkout in each submodule. We already run a checkout but typically do a sanity check first.
       In the case you aren't on the default release, you can add -f to go ahead and force the checkout,
       skipping the sanity check

amstewart:
You've got two identical commits in this patchset now. Did you mean to squash them into one?

chaitu236 (approved):

PR #213, opened 2023-01-05
create-build-nilrt.sh: Optionally accept codename on command line

mike-petersen-ni:
The meta-nilrt repository, as one of the high higher-activity repositores, will likely be the main instigator of fresh automatic builds. As it's only used here to determine the codename, specifying that independently will reduce build load.

This change leaves the script compatible with any previous usage.

This change will need to be cherry-picked to other branches. There will probably be a conflict for the academic branch, but that shouldn't be too difficult to resolve.

### DIFF
--- a/docker/create-build-nilrt.sh
+++ b/docker/create-build-nilrt.sh
@@ -9,26 +9,38 @@ IMAGE_NAME=build-nilrt
 
 usage() {
 	cat >&2 <<EOF
-$(basename) [TAG_BRANCH] [TAG_CODENAME]
+$(basename) [-c TAG_CODENAME] [TAG_BRANCH]
 Creates a new build-nilrt container based on the garmin/pyrex containers, and
 tags it with the current nilrt.git repo shorthash, as 'latest', and
 (optionally) with the current OE branch name.
+
+Arguments:
+-c, --codename  The codename (e.g. 'hardknott') of the meta-nilrt.git repo. If
+                not set, the codename will be parsed from meta-nilrt's
+                layer.conf.
 
 Positionals:
 TAG_BRANCH  The 'branch' name of the current nilrt.git repo (eg. 'dunfell'). If
             not set, the docker container will not be tagged with the branch
             name.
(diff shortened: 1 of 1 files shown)


docker/create-build-nilrt.sh
```diff
@@ -9,7 +9,7 @@ IMAGE_NAME=build-nilrt
 
 usage() {
 	cat >&2 <<EOF
-$(basename) [TAG_BRANCH]
+$(basename) [TAG_BRANCH] [TAG_CODENAME]
```
amstewart: Could you use an argument, instead of a positional for the `TAG_CODENAME`?

amstewart (changes requested):

amstewart (approved):

amstewart:
Cherry-picked to `nilrt/master/kirkstone`.

docker/create-build-nilrt.sh
```diff
@@ -45,13 +59,15 @@ pushd "${SCRIPT_ROOT}"
...
+# optionally parse the NIRLT codename from the meta-nilrt:layer.conf
+if [ -z "${tag_codename}" ]; then
+	NILRT_codename=$(grep -e '^LAYERSERIES_COMPAT_meta-nilrt' "${SCRIPT_ROOT}/../sources/meta-nilrt/conf/layer.conf" | cut -d'"' -f2)
+	if [ -z "$NILRT_codename" ]; then
+		echo "ERROR: could not parse NILRT_codename from the meta-nilrt layer." >&2
+		exit 1
+	else
+		echo "INFO: using NILRT_codename=${NILRT_codename}"
+	fi
 fi
```
chaitu236, line 71: Do we need an else condition here that sets `NILRT_codename` to `tag_codename`, or just rename `tag_codename` to `NILRT_codename`? I don't see `NILRT_codename` being initialized anywhere.

mike-petersen-ni: Ah, dangit. You're right.

PR #222, opened 2023-03-09
[kirkstone] Revert "auto.conf: Enable cve-check"

SparkingSpork:
This reverts commit 35cb7382b2b2f08516eefcd61d06446ae9142cd7.

A change upstream in kirkstone causes builds with both BUILD_IMAGES_FROM_FEEDS and cve-check to fail (see commit 291024f in openembedded-core). Instead of enabling the cve-check for all builds including images, it will need to be enabled in our internal feed build configurations.


amstewart (approved):
I'll let you pull this @SparkingSpork

PR #230, opened 2023-03-27
[23.3/hardknott] sources: Update submodule refs

mike-petersen-ni:
This is a convenience for people checking out the release branch--it prevents git from reporting changes due to the stale refs.

This could be cherry-picked into `nilrt/master/hardknott` at maintainer's discretion, if we also want those bumped to the most recent release.


amstewart (approved):
CP not needed.

PR #234, opened 2023-04-26
[hardknott] cve-to-csv.py: Update regex to explicitly look for keys.

SparkingSpork:
The regex used to parse fields in CVE files was failing when text in the description matched the key filter. This change updates the regex to only check for known key values explicitly.

[AB#2377202](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2377202)

## Testing done: 
Built packagefeed-ni-core locally with cve-check enabled, then ran the script to output the CSV. Example CSV here:
[cve.csv](https://github.com/ni/nilrt/files/11336323/cve.csv)


SparkingSpork:
Note that this will need cherry-picked into kirkstone as well.

chaitu236 (approved):

amstewart (approved):

amstewart:
Cherry-picked to kirkstone.

PR #236, opened 2023-06-14
[kirkstone] Updates to kernel readme

SparkingSpork:
After reviewing the kernel build instructions, there were several minor updates that could be made. This included:
- Updating toolchain build instructions in README.md to include Windows toolchains
- Updating the links to get toolchain downloads in the kernel readme
- Updating tone and language for troubleshooting if something doesn't boot after a kernel update
- Add ssh to commands that need to run on the target
- Update the ARM build instructions to be in line with the x64 instructions.

[AB#2308610](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/2308610)

## Testing
Ran through the complete instructions for x64 and confirmed they worked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,15 +100,6 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
 
     These scripts are also a good source for understanding the steps to build a package feed manually. Note that if you are building package feeds *manually*, you must bitbake the special `package-index` target before using the feed.
 
-    * ##### Building the cross-compile toolchain
-        One of these pipelines can build the cross-compile toolchain. By default, it builds for x86_64 Linux hosts.
-
-        ```bash
-        bash ../scripts/pipelines/build.toolchain.sh
-        ```
-
-        During the build, a script is generated at `$BUILDDIR/tmp-glibc/deploy/sdk`, with a name like `oecore-x86_64-core2-64-toolchain-9.2.sh`. The script is a self-extracting archive, and can be copied to and executed on an appropriate host system to install the toolchain.
-
 7. #### Building various images
 
     **NOTE** You must build packagefeed-ni-core and package-index first to build images.
@@ -149,6 +140,17 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
         Boot your NI Linux Real-Time compatible hardware from the recovery media and follow on-screen instructions to perform a factory reset.
 
     **NOTE** By default, National Instruments software is pulled from a feed hosted on ni.com. You can redirect to a mirror by setting IPK_NI_SUBFEED_URI to any URI supported by opkg in your org.conf,site.conf, or auto.conf.
+
+8. #### Building the cross-compile toolchain
+
+    In order to compile custom packages for NI Linux Real-Time on a host system, a cross-compile toolchain is necessary. This can be built directly
+    from one of the scripts in the [`:scripts/pipelines/`](https://github.com/ni/nilrt/tree/HEAD/scripts/pipelines) directory. By default, it builds for x86_64 Linux hosts.
+
+    ```bash
+    bash ../scripts/pipelines/build.toolchain.sh
+    ```
+
+    During the build, a script is generated at `$BUILDDIR/tmp-glibc/deploy/sdk`, with a name like `oecore-x86_64-core2-64-toolchain-9.2.sh`. The script is a self-extracting archive, and can be copied to and executed on an appropriate host system to install the toolchain.
 
 ---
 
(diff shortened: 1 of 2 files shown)


SparkingSpork:
Note that this will need cherry-picked into hardknott.

README.md
```diff
@@ -100,16 +100,16 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
 
     These scripts are also a good source for understanding the steps to build a package feed manually. Note that if you are building package feeds *manually*, you must bitbake the special `package-index` target before using the feed.
 
-    * ##### Building the cross-compile toolchain
-        One of these pipelines can build the cross-compile toolchain. By default, it builds for x86_64 Linux hosts.
+7. #### Building the cross-compile toolchain
+    One of the scripts can build the cross-compile toolchain. By default, it builds for x86_64 Linux hosts.
```
amstewart: My only issue here is that the fact that this is item 7 implies to a builder that they need to do it prior to building images, when it isn't necessary. Can we make it more clear that it is an optional step?

docs/README.kernel.md
```diff
@@ -231,10 +231,10 @@ KERNEL_VERSION=`make -s kernelrelease`
    issue), the `/boot/runmode/bzImage` path used by the bootloader is
```
amstewart, line 231: NIT: commit message a5494271e132d8c1698f822c3b7484f11a19646a

`s/hte/the/`

docs/README.kernel.md
```diff
@@ -231,10 +231,10 @@ KERNEL_VERSION=`make -s kernelrelease`
...
    not a symbolic link, but rather the kernel itself. To avoid
    overwriting the known-good kernel, rename it appropriately by running
-   this command on the target:
+   this command:
 
    ```bash
-   test ! -h /boot/runmode/bzImage && mv /boot/runmode/bzImage /boot/runmode/bzImage-`uname -r`
+   ssh admin@$TARGET "test ! -h /boot/runmode/bzImage && mv /boot/runmode/bzImage /boot/runmode/bzImage-`uname -r`"
```
amstewart, line 237: I'm not so sure about the value of giving the ssh syntax. I think we would like to maintain that clients can run these commands however they want. And it seems like it would be more annoying to have to write in their own `TARGET` hostname every time, instead of just being able to copy the whole command into whatever ssh (or serial) connection they have open.

amstewart (changes requested):

docs/README.kernel.md
SparkingSpork: We tell clients to use a `TARGET` and `KERNEL_VERSION` [here](https://github.com/ni/nilrt/blob/01444364b22aa57866b1cf66264fcb3015e4b238/docs/README.kernel.md?plain=1#L225-L228). They would still be copying without editing assuming they set those, and I would assume a client who wants to use serial or a separate ssh connection would realize they don't need the ssh portion when copying.

Additionally, some of the commands use that `KERNEL_VERSION` variable that variable wouldn't exist in the target connection unless I set it separately.

chaitu236 (approved):

docs/README.kernel.md
SparkingSpork: Fixed.

README.md
SparkingSpork: Moved it to step 8 with an explanation of when the toolchain is needed.

SparkingSpork:
### V2
- Commit message typo fixed
- Moved toolchain build step to after images to reduce confusions

docs/README.kernel.md
amstewart: Yeah; that's fine.

amstewart (approved):

README.md
```diff
@@ -150,6 +141,27 @@ This project uses the [pyrex](https://github.com/garmin/pyrex) tool to transpare
 
     **NOTE** By default, National Instruments software is pulled from a feed hosted on ni.com. You can redirect to a mirror by setting IPK_NI_SUBFEED_URI to any URI supported by opkg in your org.conf,site.conf, or auto.conf.
 
+8. #### Building the cross-compile toolchain
+
+    In order to compile custom packages for NI Linux Real-Time on a host system, a cross-compile toolchain is necessary. This can be built directlry
```
gratian: directlry -> directly

docs/README.kernel.md
```diff
@@ -368,6 +399,13 @@ tar cz --exclude=./.git --exclude=$TEMP_MODULES . | ssh admin@$TARGET tar xz --n
 
 ### Using the Source with DKMS
 
+The following steps are all run on the target. This can be done either by
```
gratian, line 402: Is there an 'or' that goes with the 'either' above? I suspect the 'or' is a 'serial console'?

gratian (approved):
In general looks good, just a couple of spelling nits.

SparkingSpork:
## V3
- Fixed spelling/language mistakes from Gratian's feeback.

PR #267, opened 2024-03-01
scripts: add host_deploy_ipks script

amstewart:
Add a developer script which automates the process of hosting an OE build's deploy/IPKs tree on the network, which makes it faster to test out IPK changes on real hardware and VMs.

```
[0] usr0:dev$ ./host_deploy_ipks -h
usage: host_deploy_ipks [-h] [--verbose] [--version] [-b [BIND]] [-e] [-p [PORT]] [--suppress-opkg-conf] [root]

Host an OE IPK deploy tree on an HTTP server.

positional arguments:
  root                  The root path for the feed server. (Default=${PWD})

options:
  -h, --help            show this help message and exit
  --verbose, -v         Increase the verbosity of script output. Specify once for verbose output, specify twice for debug.
  --version, -V         Print his script's version information and exit.
  -b [BIND], --bind [BIND]
                        The host address to which the feed server should bind. (Default=0.0.0.0)
  -e, --host-empty      If asserted, host IPK feeds that contain no packages.
  -p [PORT], --port [PORT]
                        The port number over which the feed server will host. (Default=8080)
  --suppress-opkg-conf  If enabled, do not print an opkg conf script.
```

Invocations look like:
```
[0] usr0:dev$ ./host_deploy_ipks -v ../../build/tmp-glibc/deploy/ipk/
INFO:host_deploy_ipks:Searching ../../build/tmp-glibc/deploy/ipk for IPK feeds.
INFO:host_deploy_ipks:Found IPK feed: ../../build/tmp-glibc/deploy/ipk
WARNING:host_deploy_ipks:Skipping empty index: ../../build/tmp-glibc/deploy/ipk/Packages
INFO:host_deploy_ipks:Found IPK feed: ../../build/tmp-glibc/deploy/ipk/all
INFO:host_deploy_ipks:Found IPK feed: ../../build/tmp-glibc/deploy/ipk/core2-64
INFO:host_deploy_ipks:Found IPK feed: ../../build/tmp-glibc/deploy/ipk/x64
INFO:host_deploy_ipks:Created temporary feed directory: /tmp/host_deploy_ipks.or54aurp
=== Copy+Paste to setup opkg config ===
cat >/etc/opkg/dev.conf <<-EOF
src  all  http://41d17:8080/all  [trusted=yes]
src  x64  http://41d17:8080/x64  [trusted=yes]
src  core2-64  http://41d17:8080/core2-64  [trusted=yes]
EOF
===
Serving IPKs on 0.0.0.0:8080 ...
```
After which, you can copy-paste the snippet into a target's shell and install opkg packages to your contentment. (Provided your dev machine's hostname is resolvable from the target.)

Send `^C` to stop the feed server.

# Testing
* [x] Tested the script from my dev machine with a QEMU VM. After adding my QEMU host to the guest like `echo "10.0.2.2 41d17" >>/etc/hosts` (only needed to make QEMU SLiRP networking work), I can trivially install IPKs from the hosted feed.

### DIFF
--- a/scripts/dev/host_deploy_ipks
+++ b/scripts/dev/host_deploy_ipks
@@ -7,15 +7,15 @@ from argparse import ArgumentParser, Namespace
 from glob import glob
 from http.server import HTTPServer, SimpleHTTPRequestHandler
 import logging
-from pathlib import Path
 import os
+from pathlib import Path
 import re
 from socket import getfqdn
 import sys
 from tempfile import TemporaryDirectory
 import urllib.parse
 
-__author__ = "Alex Stewart <astewart.49c6@gmail.com>"
+__author__ = "Alex Stewart <alex.stewart@ni.com>"
 __version__ = "1"
 SCRIPT_NAME = "host_deploy_ipks"  # value intentionally not computed at runtime
 
--- a/scripts/dev/host_deploy_ipks
+++ b/scripts/dev/host_deploy_ipks
@@ -34,8 +34,8 @@ class IPKFeedServer(HTTPServer):
     def __init__(
         self,
         root: str | os.PathLike,
-        bind_address: str=HTTP_DEFAULT_HOST,
-        host_port: int=HTTP_DEFAULT_PORT,
+        bind_address: str = HTTP_DEFAULT_HOST,
+        host_port: int = HTTP_DEFAULT_PORT,
     ):
         """Creates a new IPKFeedServer instance.
 
@@ -64,10 +64,10 @@ class IPKFeedDirectory(TemporaryDirectory):
 
     PREFIX = SCRIPT_NAME + "."
 
-    def __init__(self, ipk_directories: list[Path], dir: (None | os.PathLike)=None):
+    def __init__(self, ipk_directories: list[Path], dir: None | os.PathLike = None):
         """Initialize a new IPKFeedDirectory, and create a symlink farm pointing
         to the directories that will be hosted on the server.
-        
+
         Args:
             ipk_directories: A list of a ipk_directory paths to host.
             dir: If None, create the temporary directory in the TMPDIR location;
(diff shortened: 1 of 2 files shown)


scripts/dev/host_deploy_ipks
```diff
@@ -0,0 +1,236 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Host an OE IPK deploy tree on an HTTP server."""
+
+from argparse import ArgumentParser, Namespace
+from glob import glob
+from http.server import HTTPServer, SimpleHTTPRequestHandler
+import logging
+from pathlib import Path
+import os
```
chaitu236, line 10: Alphabetically should go before `from pathlib..`. Also a good idea to run https://pypi.org/project/flake8/, https://pypi.org/project/black/.

scripts/dev/host_deploy_ipks
```diff
@@ -0,0 +1,236 @@
...
+    parser.add_argument("-b", "--bind", nargs="?", default=HTTP_DEFAULT_HOST,
+        help="The host address to which the feed server should bind. (Default=%(default)s)")
+    parser.add_argument("-e", "--host-empty", action="store_true",
+        help="If asserted, host IPK feeds that contain no packages.")
+    parser.add_argument("-p", "--port", type=int, nargs="?", default=HTTP_DEFAULT_PORT,
+        help="The port number over which the feed server will host. (Default=%(default)s)")
+    parser.add_argument("--suppress-opkg-conf", action="store_true",
+        help="If enabled, do not print an opkg conf script.")
+    
+    parser.add_argument("root", type=Path, nargs="?", default=os.getcwd(),
```
chaitu236: [Optional] It'd be nice if the script can autolocate this as location of feeds is fixed wrt location of the script.

chaitu236 (approved):

scripts/dev/host_deploy_ipks
amstewart: Will fix.

scripts/dev/host_deploy_ipks
amstewart: That would mean that calling the script with no arguments would start hosting content on the network. I don't want that to be the case as a safety for if someone accidentally tab-completes or dmenu's the script.

I could add a `--auto` argument or something to try and auto-locate the deploy IPKs. But at that point, you may as well instead pass the path yourself.

So I'm leaning toward not doing this.

PR #271, opened 2024-04-02
[24.3/kirkstone] docs: update feed changelogs

amstewart:
Service for the 24.3 release.

# Testing
* [x] Visual check only; non-functional changes.

# Meta
* Will cherry-pick to `nilrt/master/kirkstone` and `nilrt/master/next`.


chaitu236:
I'm ok with merging this only if this would **not** trigger repohashes pipeline (and subsequently a build). `update-from-github-v2.yml` seems to suggest repohashes pipeline won't run but I'd like someone else to also confirm.

@amstewart @jeminor

jeminor:
> I'm ok with merging this only if this would **not** trigger repohashes pipeline (and subsequently a build). `update-from-github-v2.yml` seems to suggest repohashes pipeline won't run but I'd like someone else to also confirm.
> 
> @amstewart @jeminor

Confirmed, but I'm also ok with not making the change in the release branch for this release given our discussion around if this is worth keeping around at all.

amstewart:
Yeah. My expectation was that this wouldn't trigger a build or go into finals. And my intention was to merge this and them just add a conversation topic to the huddle to discuss dropping these files and our process around generating them.

chaitu236 (approved):

PR #278, opened 2024-06-20
Update README.md

pratheekshasn:
Change the bitbake command for installing python to specify python3.

Currently, running the command `bitbake python ruby apache2` gives the following error:

```
ERROR: Nothing PROVIDES 'python'. Close matches:

  python3
  python3-py
  python3-sh
```

Changing it to `bitbake python3 ruby apache2`


chaitu236 (changes requested):
Change looks ok but there is no work item number in the description.

I'm going to merge this in the interest of time but please update the description afterwards.

PR #280, opened 2024-08-20
build.core-feeds.sh: Update to build Kernel

usercw88:
Update build.core-feeds.sh script to additionally build Linux Kernel in order to generate CVE

### DIFF
--- a/scripts/pipelines/build.core-feeds.sh
+++ b/scripts/pipelines/build.core-feeds.sh
@@ -7,3 +7,6 @@ SCRIPT_ROOT=$(realpath $(dirname $BASH_SOURCE))
 echo "INFO: Building the core package feed."
 bitbake packagefeed-ni-core
 bitbake package-index
+
+# AB#2791248: Build the Linux Kernel to ensure CVEs are made available
+bitbake linux-nilrt


scripts/pipelines/build.core-feeds.sh
```diff
@@ -7,3 +7,12 @@ SCRIPT_ROOT=$(realpath $(dirname $BASH_SOURCE))
...
 bitbake package-index
+
+# AB#2791248: Build the Linux Kernel to ensure CVEs are made available
+if [ "$1" == "next" ]; then
+   echo "INFO: Building the Linux Kernel Next"
+   bitbake linux-nilrt-next
+else
+   echo "INFO: Building the Linux Kernel"
+   bitbake linux-nilrt
+fi
```
amstewart: Why the conditionals? It shouldn't harm anything to ask bitbake rebuild them each time; right? It will simply skip everything in sstate.

amstewart (changes requested):

scripts/pipelines/build.core-feeds.sh
usercw88: The `build.core-feeds.sh` script is currently executed as a part of `oe_feed_core` and `oe_feed_core_next`. The goal was to only build the relevant kernel for each feed.
Additionally, I wasn't sure we wanted the CVEs for linux-nilrt-next to be exported for `oe_feed_core` as it might have an impact on release

amstewart: I think `linux-nilrt-next` and `nilrt-next` are decoupled components. The -next _distribution_ doesn't always use the -next kernel. In fact, I think it never does. If the `linux-nilrt-next` kernel is currently in the extra/ packagefeed, then we don't need to have it in this pipeline script in any case, and can just build `linux-nilrt` for both the -next and current distribution components.

amstewart (approved):

amstewart:
@chaitu236 Could you pull this into the kirkstone and master refs?

chaitu236:
> @chaitu236 Could you pull this into the kirkstone and master refs?

It is not clear to me if this addresses the root cause of the issue or is just a short term workaround for kernel recipe - seems more like the latter. Have we determined if there are other recipes that could be affected by this issue?

I don't mind pulling this if we intend this as a short term workaround but we should address the underlying issue or at least make sure no other recipes are affected.

@amstewart @usercw88 @gratian

usercw88:
@chaitu236 This change is intended to be a workaround in order to retrieve CVEs for the kernel. More investigation is needed to determine why it isn't being generated when a part of the packagefeed-ni-core packagegroup

cc @amstewart @gratian

chaitu236:
Cherry-picked to nilrt/master/next - 59ebeff6420dde2237b36ffffaefc8727172ffeb

PR #282, opened 2024-09-04
repos.conf: Updated upstream sources for scarthgap

erickshepherdNI:
More repos now have scarthgap branches so update those.

[AB#2755018](https://dev.azure.com/ni/DevCentral/_workitems/edit/2755018)


chaitu236 (changes requested):
Changes are ok but commit message is misleding.

I understand you copied the message from when Alex made changes to this file but at that point we were doing a rebase of NI commits on top of upstream master branches; but that isn't the case at this point.

Something like `upstream_merge: Update some repos to scarthgap` would be more accurate.

chaitu236 (approved):

PR #283, opened 2024-09-09
Revert "build.cross-toolchain.sh: Set SDK_ARCHIVE_TYPE to "zip""

usercw88:
This reverts commit 7ea0563d37b89d6a03cd60e6ad740d3a5b28dfa3.

The cross-toolchain contains several symlinks that become "broken" when archiving as a zip

[WI#2849157](https://dev.azure.com/ni/DevCentral/_workitems/edit/2849157)


chaitu236 (approved):
Changes look ok to me - assuming azdo code is being changed as well. Let me know when this is ready to be merged.

chaitu236 (changes requested):
Sorry I just unzipped the latest .zip from an export on a Linux machine and I do see symlinks in it. The `as.exe` symlink mentioned in https://dev.azure.com/ni/DevCentral/_workitems/edit/2849157 (that I assume prompted this revert) also exists and correctly points to `x86_64-nilrt-linux-as.exe`.

I'm not sure if "zip" format is the cause of broken symlinks in installer. Or am I missing something?

usercw88:
The issue is with unzipping on a Windows system. Depending on the tool used, the symlinks become (broken) applications. 7zip has the capability to unzip these with symlinks, but it requires elevated/administrative privileges. Given the temperamental relationship Windows has with symlinks, I think it's best to keep the toolchain as a tar until we need to use it on a Windows system

chaitu236:
> The issue is with unzipping on a Windows system. Depending on the tool used, the symlinks become (broken) applications. 7zip has the capability to unzip these with symlinks, but it requires elevated/administrative privileges. Given the temperamental relationship Windows has with symlinks, I think it's best to keep the toolchain as a tar until we need to use it on a Windows system

Don't we extract this in a build machine to create installers? In which case can't we just use any tool we like to do the extraction or extract with admin privileges?

usercw88:
> Don't we extract this in a build machine to create installers? In which case can't we just use any tool we like to do the extraction or extract with admin privileges?

For clarification, on a **Windows system**, I have not been able to extract the contents of the .zip with the symlinks intact. Whether with or without admin privileges, via the CLI or GUI, the symlinks always come out as applications.
On a Windows system, using 7z with admin privileges, via the CLI, and using the .tar.xz, it is possible to extract the symlinks.
Note: On my local build machine, I have not been able to get 7z to "run as administrator", leading to errors when handling the symlinks

> I just unzipped the latest .zip from an export on a Linux machine and I do see symlinks in it

I experience the same thing, Linux/tar is able to maintain the symlinks whether from a .zip or .tar

Given that the solution for the windows toolchain installer is rooted on a Linux build machine, the use of a .zip or .tar is not crucial. The purpose of reverting this back to a tar is, more or less, to ensure that those who obtain the .zip on a windows machine can correctly extract it

chaitu236 (approved):

chaitu236:
Didn't notice this before but please update PR description with WI#.

chaitu236:
Cherry-picked to nilrt/master/next 919a4ad4bcd4e0eb7c73969c9d9811729caa8675

PR #285, opened 2024-09-11
Update build-nilrt.Dockerfile to enable Talladega to use the exported docker images

jtongen1:
Talladega would like to use the nilrt-built docker images in `build_nilrt_containers` to build Talladega images.  The Talladega build process requires the 2 components added in this PR to be included in the docker image our our build will fail.

### DIFF
--- a/docker/build-nilrt.Dockerfile
+++ b/docker/build-nilrt.Dockerfile
@@ -6,6 +6,8 @@ RUN apt-get update && apt-get install --assume-yes \
 	genisoimage \
 	qemu-system-x86 \
 	qemu-utils \
+	libtinfo5 \
+	openssh-client \
 ""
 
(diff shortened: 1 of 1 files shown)


jtongen1:
It looks like the .Dockerfile that is modified in this PR is used to build those docker images (and not the files in branch `nilrt/<version>/kirkstone` but if I've got that wrong please let me know.

chaitu236 (changes requested):
Changes look fine but can you sign the commit. And also break the second line of commit message into multiple lines so each one is no longer than 72 chars.

If there is a work item related to this change, please mention it in PR description.

chaitu236:
Cherry-picked to nilrt/master/next 7fa27551d853ce3fe25edb97a5c283f2a48eacaf

PR #308, opened 2025-06-04
cve-to-csv-v2: Create script using json exports

usercw88:
# POC: Not intended for submission

# Changes

Created `cve-to-csv-v2.py` to use json export files to generate CVE Summary

# Testing

Tested script against `rtos_oe_feeds 25.5d81` oe-core cve folder
  - Validated generated `cve_summary.csv` against exported `cve_summary.csv`
    - Equivalent number of CVEs generated

# Process

* [ ] This PR should be cherry-picked to the [`next/`](https://github.com/ni/nilrt/tree/nilrt/master/next) ref.

__Suggested Reviewers:__
* @ni/rtos

### DIFF
--- a/scripts/pipelines/cve-to-csv-v2.py
+++ b/scripts/pipelines/cve-to-csv-v2.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+from argparse import ArgumentParser
+from pathlib import Path
+import json
+import os
+import sys
+
+
+parser = ArgumentParser(
+    description="Parses a directory full of OE cve_check json files, and writes their csv equivalent to stdout.",
+    )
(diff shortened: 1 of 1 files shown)


chaitu236 (commented):
How is this different from https://github.com/ni/nilrt/blob/nilrt/master/scarthgap/scripts/pipelines/cve-to-csv.py and why is this required?

usercw88:
> How is this different from https://github.com/ni/nilrt/blob/nilrt/master/scarthgap/scripts/pipelines/cve-to-csv.py and why is this required?

The `cve-to-csv.py` uses files with specific formatting, and can break if new sections are added (for example scorev4 caused it to fail).

PR #311, opened 2025-06-23
Automating upstream merge and testing

CommitSentience:
# Justification
[AB#3174362](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3174362)

Currently, upstream merge is handled by running [this](https://github.com/ni/nilrt/blob/nilrt/master/hardknott/scripts/dev/upstream_merge/upstream_merge.sh) shell script. Although the script is perfectly capable of merging new commits to the existing ni forks, testing the new changes and creating PRs for the forks of each submodule is still a manual process. The current PR takes care of automating this step as well.

# Implementation
This automation streamlines the process of merging upstream changes into the nilrt repository, building core images, and testing them on a VM, and automates PR creation and email notifications. It reduces manual effort, ensures the fork stays up-to-date.

The original shell script has now been rewritten as a python script such that it can be reused by other scripts we may create in the future.

# Testing
These changes have been tested on my fork by running the full automation workflow, including merge, build, test, and PR creation.

[This](https://github.com/pratheekshasn/nilrt/pull/2) PR was created by running the push_branch_and_create_pr script.

### DIFF
--- a/scripts/dev/upstream_merge/README.md
+++ b/scripts/dev/upstream_merge/README.md
@@ -1,0 +1,229 @@
+# MOTIVATION
+
+To ensure the long-term maintainability and efficiency of the ni/nilrt repository, automating the upstream merge process is a crucial step. Manual upstream merges are often tedious, time-consuming, and prone to human error, especially as the size and frequency of upstream changes grow. By automating this process, we can significantly reduce the cost of maintenance by saving several hours of manual effort per release cycle. This allows developers to focus more on value-adding tasks rather than routine integration work. Automation also ensures that our fork stays closely aligned with upstream changes, minimizing future integration conflicts and making it easier to adopt new features, security patches, and bug fixes promptly. Overall, this improves the quality, stability, and security of the codebase while enhancing the team's productivity and responsiveness to upstream evolution.
+
+The script does the following tasks:
+- Automatically merges changes from upstream repositories into the local fork, reducing manual effort.
+- Builds the safemode and runmode images after a successful merge to ensure the codebase is functional.
+- Installs and tests these images on a RT target via SSH to verify that the integration is successful and the target works as expected.
+- Commits and pushes the new changes and creates a PR targeting the main branch.
+- Sends email reports about the merge and build status.
+
+Overall, it streamlines and safeguards the process of keeping the local fork up-to-date with upstream changes, ensuring quality and reducing maintenance overhead.
+
+---
+
+## **Script Overview**
+
+### **Workflow for Each Submodule in the `nilrt` Repository**
+
+1. **Check Out the Base Branch**:
+   - The script checks out the base branch as specified in the [`repos.conf`](./repos.conf) file.
+
+2. **Pull the Upstream Branch**:
+   - The upstream branch mentioned in [`repos.conf`](./repos.conf) is pulled to ensure the latest changes are fetched.
+
+3. **Merge the Two Branches**:
+   - The base branch and the upstream branch are merged.
+   - The script reports whether the merge was successful or if it encountered any conflicts.
+
+4. **Build Images on Successful Merge**:
+   - If the merge is successful, the script proceeds to build the following images:
+     - **Safemode Image**
+     - **Runmode Image**
+
+5. **Install Images on the RT Target**:
+   - The safemode and runmode images are installed on a RT target via SSH.
+   - The script then verifies the installation by running tests on the RT target (such as checking the OS version ).
+
+6. **Push Changes and Create Pull Requests**:
+   - If the build and test steps succeed, the script commits and pushes the new changes to the downstream fork repository and automatically creates a PR targeting the main branch.
+
+---
+
+## **Configuration File: `automation_conf.json`**
+
+The script is user configurable and uses `automation_conf.json` to define various parameters. **All fields are required**, and if a field is not present, `None` will be used as the default value. Below is an explanation of the fields in the configuration file:
+
+- **`nilrt_branch`**:
+  The nilrt branch to pull the latest changes from.
+
+- **`meta_nilrt_branch`**:
+  The meta-nilrt branch to pull the latest changes from.
+
+- **`conf_file_path`**:
+  Path to the configuration file (default: [`repos.conf`](./repos.conf)).
+
+- **`force_checkout`**:
+  Enables forceful checkout to the base branch if set to `True`.
+
+- **`upstream_repo_name`**:
+  The name of the upstream remote repository.
+
+- **`merge_branch_name`**:
+  The name of the branch where the merge will be performed.
+
+- **`username`**:
+  GitHub username where the forks are maintained.
+
+- **`fork_name`**:
+  The name of the downstream fork repository where the merge will be performed. This is the repository owned by the user specified in the username field.
+
+- **`email_from`**:
+  The email address from which the merge report will be sent.
+
+- **`email_to`**:
+  The email address to which the merge report will be sent.
+
+- **`email_log_level`**:
+  - **`0`**:
+    - Includes the status of the upstream merge:
+      - **`... OK`**: Merge completed successfully.
+      - **`... OK (no changes)`**: No changes were detected during the merge.
+      - **`... ERRORS`**: Errors occurred during the merge.
+    - In case of a merge conflict, the error details will also be included in the email.
+
+  - **`1`**:
+    - Includes the diff (differences) for a successful merge in addition to the status.
+
+- **`log_level`**:
+  Integer representing the logging level:
+  - `10`: DEBUG
+  - `20`: INFO
+  - `30`: WARNING
+  - `40`: ERROR
+  - `50`: CRITICAL
+
+- **`build_args`**:
+  String of extra arguments to pass to the build process (e.g., `"--org"` for NI corporate network builds).
+
+- **`rt_target_IP`**:
+  The IP address or hostname of the RT target where images will be installed and tested via SSH.
+
+**`Note`**:
+- If the configuration file is `automation_conf.json`, you do not need to specify its path explicitly, as it is set as default. However, if you are using a different configuration file, you must provide its path using the `-c` argument.
+For example:
+
+```bash
+python3 scripts/dev/upstream_merge/upstream_merge_and_test.py -c path/to/your_config.json -w workItemID
+```
+
+
+---
+
+# HOW TO USE THE MERGE SCRIPT
+
+### **1. Navigate to the root of the nilrt repo**
+```bash
+cd ~/nilrt
+```
+
+### **2. Run the Script**
+#### **To Perform a Merge**
+This will merge upstream changes, build the images, test them on the RT target, and create PRs with the new changes after the build and test steps succeed.
+
+Note that these steps are serial. The build should succeed for the test to run, and the test should pass for the PRs to be created.
+
+```bash
+python3 scripts/dev/upstream_merge/upstream_merge_and_test.py -c path/to/your_config.json -w workItemID
+```
+
+- **`workItemID`**:
+The `workItemID` associated with the pull request.
+
+
+#### **To Skip the Merge and Only Build, Test, and Create PRs**
+
+Enable this option when a merge conflict has occurred and you have resolved it manually.
+This option is used to skip the automated upstream merge step in the pipeline, allowing the build and test stages to proceed with your manually resolved state.
+
+**How to Manually Resolve a Merge Conflict:**
+
+Once you have resolved the conflict and committed the changes on the build machine, you can rerun the script with the `-s` option to skip the automated merge and proceed directly to the build and test stages.
+
+```bash
+python3 scripts/dev/upstream_merge/upstream_merge_and_test.py -c path/to/your_config.json -w workItemID -s
+```
+#### **To Skip Pushing the Changes and Creating PRs**
+
+Enable this option if you want to run the merge, build, and test steps, but do **not** want to push any changes or create pull requests. This can be useful for dry runs, debugging, or when you want to verify the process without affecting the remote repository.
+
+To use this option, pass the `-skip_push_and_pr` flag when running the script:
+
+```bash
+python3 scripts/dev/upstream_merge/upstream_merge_and_test.py -c path/to/your_config.json -w workItemID -skip_push_and_pr
+```
+
+---
+
+## **Supporting Files**
+
+### **`shell_commands.py`**
+- Provides utility functions to execute shell commands.
+- Acts as a wrapper for running system commands from Python.
+
+### **`git_commands.py`**
+- Contains various Git-related functions that rely on `shell_commands.py`.
+- Includes operations like fetching, pulling, creating branches, and merging.
+
+### **`git_repo.py`**
+- Defines the `git_repo` class, which encapsulates details and operations for a Git repository.
+-  Data members are:
+    - local_repo
+    - local_base_branch
+    - upstream_branch
+    - upstream_repo_name
+    - upstream_repo_url
+    - fork_name
+    - fork_url
+
+### **`upstream_merge.py`**
+- Contains helper functions for merging submodules with their respective upstream repositories.
+
+### **`build.py`**
+- Handles the process of building images.
+- Key steps include:
+  - Setting up the Docker environment.
+  - Building core feeds.
+  - Building core images.
+
+### **`test.py`**
+- Manages testing of the built images on an RT target via SSH.
+- Key steps include:
+  - Installing and testing safemode and runmode images.
+  - Verifying the OS version.
+
+### **`push_branch_and_create_pr.py`**
+- Handles pushing the merged changes to the downstream fork repository.
+- Automates the creation of a pull request on GitHub after a successful merge and build.
+- Uses the GitHub CLI or API to generate PRs with appropriate titles, descriptions, and checklists.
+
+> **Important Note:**
+> The `push_branch_and_create_pr.py` script performs a **force push** (`git push --force`) to the downstream fork repository.
+> This will overwrite the remote branch history and may cause loss of commits if others have pushed to the same branch.
+> Use with caution and ensure you coordinate with your team before running this script.
+
+### **`json_config.py`**
+- Handles reading and validating the configuration file.
+- Key responsibilities include:
+  - Parsing the JSON configuration file.
+  - Validating required fields and their values.
+  - Providing easy access to configuration parameters for other scripts.
+
+### **`log_and_email_utils.py`**
+- Provides utilities for logging, formatting merge/build/test reports, and sending email notifications.
+- Used by automation scripts to set up logging, format status reports, and send summary emails after merges, builds, and tests.
+
+---
+
+## Contact
+
+If you have any questions or need assistance, feel free to contact:
+
+- Name: Shreejit C
+  - GitHub Username: Shreejit-03
+  - Email: shreejit.c@emerson.com
+
+- Name: Pratheeksha S N
+  - GitHub Username: pratheekshasn
+  - Email: pratheeksha.s.n@emerson.com
--- a/scripts/dev/upstream_merge/automation_conf.json
+++ b/scripts/dev/upstream_merge/automation_conf.json
@@ -1,0 +1,16 @@
+{
+    "nilrt_branch":"nilrt/master/scarthgap",
+    "meta_nilrt_branch":"nilrt/master/scarthgap",
+    "conf_file_path": "scripts/dev/upstream_merge/repos.conf",
+    "force_checkout": false,
+    "upstream_repo_name": "automerge_upstream",
+    "merge_branch_name": "dev/automerge/ni",
+    "fork_name": "myfork",
+    "email_log_level": 1,
+    "log_level": 10,
+    "email_from": "shreejit.c@emerson.com",
+    "email_to": "pratheeksha.s.n@emerson.com",
+    "username": "",
+    "build_args":"",
+    "rt_target_IP":""
+}
--- a/scripts/dev/upstream_merge/json_config.py
+++ b/scripts/dev/upstream_merge/json_config.py
@@ -1,0 +1,29 @@
+"""Provides utilities for reading and validating the automation_conf.json
+    configuration file."""
+
+import json
+import os
+
+
+class JsonConfig:
+    """Handles loading and validating the automation_conf.json
+        configuration file."""
+
+    def __init__(self, automation_conf_path, work_item_id):
+        with open(automation_conf_path, "r", encoding="utf-8") as file:
+            config = json.load(file)
+        self.work_item_id = work_item_id
+        self.nilrt_branch = config.get("nilrt_branch")
+        self.meta_nilrt_branch = config.get("meta_nilrt_branch")
+        self.conf_file_path = os.getcwd() + f"/{config.get('conf_file_path')}"
+        self.force_checkout = config.get("force_checkout")
+        self.upstream_repo_name = config.get("upstream_repo_name")
+        self.merge_branch_name = config.get("merge_branch_name")
+        self.username = config.get("username")
+        self.fork_name = config.get("fork_name")
+        self.email_from = config.get("email_from")
+        self.email_to = config.get("email_to")
+        self.email_log_level = config.get("email_log_level")
+        self.log_level = config.get("log_level")
+        self.build_args = config.get("build_args", "")
+        self.rt_target_IP = config.get("rt_target_IP")
--- a/scripts/dev/upstream_merge/log_and_email_utils.py
+++ b/scripts/dev/upstream_merge/log_and_email_utils.py
@@ -1,0 +1,194 @@
+"""
+Utilities for logging, formatting merge reports, and sending email
+notifications.
+"""
+
+import os
+import datetime
+import logging
+from utils.git_commands import send_email
+
+
+def setup_logging(log_level=10):
+    """
+    Set up logging configuration.
+    :param log_level: Logging level (default: 10).
+    """
+    global LOG_DIR
+    LOG_DIR = f"temp/{datetime.datetime.now().strftime('%d-%m-%Y')}"
+    os.makedirs(LOG_DIR, exist_ok=True)
+    log_file = (
+        f"{LOG_DIR}/upstream_merge_"
+        f"{datetime.datetime.now().strftime('%H-%M-%S')}.log"
+    )
+    logging.basicConfig(
+        level=log_level,
+        format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+        handlers=[logging.FileHandler(log_file, mode="a")],
+    )
+
+
+def write_email_headers(email_file_name, email_from,  email_to):
+    """
+    Write email header information to a log file.
+
+    :param email_file_name: The name of the log file to write to.
+    :param email_from: The sender's email address.
+    :param email_to: The recipient's email address.
+    """
+    with open(email_file_name, "w", encoding="utf-8") as log:
+        log.write(f"From: {email_from}\n")
+        log.write(f"To: {email_to}\n")
+        log.write("Subject: Merge Details\n\n")
+
+
+def format_status(status, message):
+    """
+    Format the status and message for display/logging.
+
+    :param status: Status code (0 for success, non-zero for errors).
+    :param message: Associated message or details.
+    :return: Tuple of formatted status strings for summary and detailed logs.
+    """
+    error_msg = f" ... ERRORS\n    {message or ''}\n\n\n"
+    ok_no_changes = " ... OK (no changes)"
+    ok_msg = f" ... OK\n    {message}\n\n\n"
+
+    if status != 0:
+        return (" ... ERRORS", error_msg)
+
+    if message is None:
+        return (ok_no_changes, f"{ok_no_changes}\n\n\n")
+
+    return (" ... OK", ok_msg)
+
+
+def format_merge_report(merge_report, email_log_level, skip_push_and_pr=False):
+    """
+    Format the merge report for email or logging.
+
+    The formatted report string is structured as follows:
+    - The first line contains the repository name.
+    - The second line contains the status of the merge (OK, ERRORS, or no
+      changes).
+    - The diff is added if there are any changes and if email_log_level is set
+      to 1.
+
+    Merge completed successfully:
+    sources/bitbake
+     ... OK
+         Push and PR ... OK
+
+    No changes were detected during the merge:
+    sources/bitbake
+     ... OK (no changes)
+
+    Errors occurred during the merge:
+    sources/bitbake
+     ... ERRORS
+    """
+    min_detail = ""
+    error_detail = ""
+    diff_detail = ""
+
+    build_and_test_detail = merge_report.pop("Build and Test")
+    if build_and_test_detail[0] == 0 and not skip_push_and_pr:
+        push_and_pr_details = merge_report.pop("Push and PR")
+    # Remove the build/test and PR results from the merge report so that
+    # the subsequent loop can focus solely on per-repository merge outcomes.
+    # This streamlines the logic and ensures only relevant entries are
+    # processed.
+
+    for git_obj, (status, message) in merge_report.items():
+        min_line, additional_line = format_status(status, message)
+
+        layer_name = git_obj.local_repo.split("/")[-1]
+        min_detail += f"{layer_name}\n{min_line}\n"
+
+        if status != 0:
+            error_detail += f"{layer_name}\n{message}\n"
+
+        diff_detail += f"{layer_name}\n{additional_line}"
+
+        if (
+            build_and_test_detail[0] == 0
+            and not skip_push_and_pr
+            and status == 0
+            and message is not None
+        ):
+            git_obj_push_details = push_and_pr_details[git_obj]
+            if git_obj_push_details[0] == 0:
+                min_detail += "     Push and PR ... OK\n"
+                diff_detail += "     Push and PR ... OK\n"
+            else:
+                min_detail += "     Push and PR ... ERRORS\n"
+                error_detail += (
+                    f"{git_obj.local_repo}\n{min_line}\n"
+                    f"     Push and PR ... ERRORS\n"
+                    f"    {git_obj_push_details[1]}\n"
+                )
+                diff_detail += (
+                    "     Push and PR ... ERRORS\n"
+                    f"    {git_obj_push_details[1]}\n"
+                )
+
+    min_detail += "Build and Test\n"
+    if build_and_test_detail[0] == 0:
+        min_detail += (
+            f"\nBuild and Test\n ... OK\n    {build_and_test_detail[1]}\n"
+        )
+        diff_detail += (
+            f"\nBuild and Test\n ... OK\n    {build_and_test_detail[1]}\n"
+        )
+    else:
+        min_detail += " ... ERRORS\n"
+        error_detail += (
+            f"\nBuild and Test\n ... ERRORS\n    {build_and_test_detail[1]}\n"
+        )
+        diff_detail += (
+            f"\nBuild and Test\n ... ERRORS\n    {build_and_test_detail[1]}\n"
+        )
+
+    if email_log_level == 0:
+        return min_detail + "\n\n" + error_detail
+    return min_detail + "\n\n" + error_detail + "\n\n" + diff_detail
+
+
+def write_log(email_file_name, contents):
+    """
+    Append the given contents to the specified log file.
+
+    :param email_file_name: The name of the log file to write to.
+    :param contents: The string content to append to the log file.
+    """
+    with open(email_file_name, "a", encoding="utf-8") as log:
+        log.write(contents)
+
+
+def write_log_and_send_email(
+    email_from, email_to, merge_report, email_log_level, skip_push_and_pr=False
+):
+    """
+    Write the merge report to a log file, then send it as an email.
+
+    :param email_from: The sender's email address.
+    :param email_to: The recipient's email address.
+    :param merge_report: The merge report data structure.
+    :param email_log_level: The log verbosity level for the email.
+    :param skip_push_and_pr: Flag to handle the merge report for skipping push
+        and PR creation.
+    """
+    email_file_name = (
+        LOG_DIR + "/"
+        f"upstream_merge_{datetime.datetime.now().strftime('%H-%M-%S')}.txt"
+    )
+    formatted_report_string = format_merge_report(
+        merge_report, email_log_level, skip_push_and_pr
+    )
+    write_email_headers(email_file_name, email_from, email_to)
+    write_log(email_file_name, formatted_report_string)
+    send_email(
+        to_address=email_to,
+        subject="Merge Details",
+        file=email_file_name
+    )
--- a/scripts/dev/upstream_merge/upstream_merge.py
+++ b/scripts/dev/upstream_merge/upstream_merge.py
@@ -1,0 +1,155 @@
+"""
+Workflow for automating the upstream merge process for submodules and
+repositories.
+"""
+
+
+def merge_upstream(git_obj, force_checkout, merge_branch_name):
+    """
+    Merge the latest changes from the upstream branch.
+
+    :param git_obj: The GitRepo object to interact with the repository.
+    :param force_checkout: Boolean indicating whether to force checkout to
+        the base branch.
+    :param merge_branch_name: The name of the merge branch to use.
+    :return: A tuple (status_code, message). Returns (0, None) if no changes or
+        (0, diff) if merged.
+    """
+    merge_prepare_details = prepare_for_merge(git_obj, force_checkout)
+    if merge_prepare_details[0] != 0:
+        return merge_prepare_details
+
+    repo_details = create_merge_branch(git_obj, merge_branch_name)
+    if repo_details[0] != 0:
+        return repo_details
+
+    commit_before_merge = git_obj.get_current_commit()
+
+    merge_result = git_obj.merge_branch(
+        f"{git_obj.upstream_repo_name}/{git_obj.upstream_branch}",
+        "Merge latest upstream",
+    )
+
+    if merge_result[0] == 0:
+        diff_output = git_obj.diff()
+        if (git_obj.get_current_commit() == commit_before_merge) or \
+                diff_output == (0, "",):
+            return (0, None)  # No changes after merge [... OK (no changes)]
+        return (0, diff_output[1])  # Changes were merged successfully [... OK]
+
+    return (1, merge_result[1])  # Merge failed [... ERRORS]
+
+
+def prepare_for_merge(git_obj, force_checkout):
+    """
+    Prepare the repository for merging by switching to the base branch, pulling
+    the latest changes, fetching from upstream, and creating the merge branch.
+
+    :param git_obj: The GitRepo object to interact with the repository.
+    :param force_checkout: Boolean indicating whether to force checkout
+        the base branch.
+    :return: A tuple (status_code, message). Returns (0, None) on success.
+    """
+
+    base_branch_details = switch_to_base_branch_and_pull(
+        git_obj,
+        force_checkout
+        )
+    if base_branch_details[0] != 0:
+        return base_branch_details
+
+    fetch_details = fetch_upstream(git_obj)
+    if fetch_details[0] != 0:
+        return fetch_details
+
+    return (0, None)
+
+
+def switch_to_base_branch_and_pull(git_obj, force_checkout):
+    """
+    Switch to the base branch, optionally force checkout,
+    set the origin remote, and pull the latest changes.
+
+    :param git_obj: The GitRepo object to interact with the repository.
+    :param force_checkout: Boolean indicating whether to force checkout
+        the base branch.
+    :return: A tuple (status_code, message). Returns (0, None) on success.
+    """
+    if not force_checkout:
+        branch_details = git_obj.branch_exists(git_obj.local_base_branch)
+        if not branch_details:
+            print(
+                f"\n    Branch {git_obj.local_base_branch} does not exist. "
+                "Exiting"
+            )
+            return (
+                1,
+                f"\n    Branch {git_obj.local_base_branch} does not exist. "
+                "Exiting",
+            )
+
+    checkout_details = git_obj.checkout_branch(
+        git_obj.local_base_branch, force_checkout=force_checkout
+    )
+    if checkout_details[0] != 0:
+        print(checkout_details[1])
+        return checkout_details
+
+    set_origin_details = git_obj.add_remote(
+        "origin",
+        f"https://github.com/ni/{git_obj.local_repo.split('/')[-1]}.git"
+    )
+    if set_origin_details[0] != 0:
+        print(set_origin_details[1])
+        return set_origin_details
+
+    pull_details = git_obj.pull(
+        branch_name=git_obj.local_base_branch, upstream_repo_name="origin"
+    )
+    if pull_details[0] != 0:
+        print(pull_details[1])
+        return pull_details
+
+    return (0, None)
+
+
+def fetch_upstream(git_obj):
+    """
+    Add the upstream remote and fetch the latest changes from the upstream
+    repository.
+
+    :param git_obj: The GitRepo object to interact with the repository.
+    :return: A tuple (status_code, message). Returns (0, None) on success.
+    """
+    add_remote_details = git_obj.add_remote()
+    if add_remote_details[0] != 0:
+        print(add_remote_details[1])
+        return add_remote_details
+
+    fetch_details = git_obj.fetch_branch()
+    if fetch_details[0] != 0:
+        print(fetch_details[1])
+        return fetch_details
+
+    return (0, None)
+
+
+def create_merge_branch(git_obj, merge_branch_name):
+    """
+    Create and check out a new merge branch, deleting it first if it already
+    exists.
+
+    :param git_obj: The GitRepo object to interact with the repository.
+    :param merge_branch_name: The name of the merge branch to create.
+    :return: A tuple (status_code, message). Returns (0, None) on success.
+    """
+    if git_obj.branch_exists(merge_branch_name):
+        git_obj.checkout_branch(git_obj.local_base_branch)
+        git_obj.delete_branch(merge_branch_name)
+
+    checkout_details = git_obj.checkout_branch(merge_branch_name, create=True)
+    if checkout_details[0] != 0:
+        print(checkout_details[1])
+        return checkout_details
+
+    return (0, None)
--- a/scripts/dev/upstream_merge/upstream_merge_and_test.py
+++ b/scripts/dev/upstream_merge/upstream_merge_and_test.py
@@ -1,0 +1,322 @@
+"""Main script for automating the upstream merge, build, test, and
+notification workflow."""
+
+import os
+import argparse
+
+from utils.test import install_and_test_image
+from json_config import JsonConfig
+from utils.git_repo import GitRepo
+from upstream_merge import merge_upstream
+from utils.push_branch_and_create_pr import push_branch_and_create_pr
+from utils.build import setup_env_and_build_packages
+from log_and_email_utils import setup_logging, write_log_and_send_email
+
+NILRT_URL = (
+    "https://github.com/ni/nilrt.git"
+)
+META_NILRT_URL = (
+    "https://github.com/ni/meta-nilrt.git"
+)
+
+
+def parse_args():
+    """
+    Parse command-line arguments for the automated repository merging script.
+
+    :return: Parsed arguments as returned by
+            argparse.ArgumentParser.parse_args().
+    """
+    parser = argparse.ArgumentParser(
+        description="Automated repository merging script"
+    )
+    parser.add_argument(
+        "-c",
+        type=str,
+        help="Path to configuration file",
+        default="scripts/dev/upstream_merge/automation_conf.json",
+    )
+    parser.add_argument(
+        "-w",
+        type=str,
+        help="Work item ID",
+        default=None,
+    )
+    parser.add_argument(
+        "-s",
+        action="store_true",
+        help="Skip merging with upstream"
+    )
+    parser.add_argument(
+        "-skip_push_and_pr",
+        action="store_true",
+        help="Skip pushing and creating PRs"
+    )
+    args = parser.parse_args()
+    return args
+
+
+def merge_submodules_with_upstream(
+    conf_file_path,
+    force_checkout,
+    username,
+    upstream_repo_name,
+    merge_branch_name,
+    fork_name,
+    skip_merge,
+):
+    """
+    Merge submodules with their respective upstream repositories
+    as defined in the repos.conf file.
+
+    :param conf_file_path: Path to the configuration file (repos.conf).
+    :param force_checkout: Boolean indicating whether to force checkout
+        to the base branch.
+    :param username: GitHub username owning the downstream fork.
+    :param upstream_repo_name: Name of the upstream remote repository.
+    :param merge_branch_name: The name of the merge branch to use.
+    :param fork_name: The name of the downstream fork repository.
+    :param skip_merge: If "True", skip the merge step.
+    :return: A dictionary mapping GitRepo objects to (status_code, message).
+    """
+    merge_report = {}
+
+    with open(os.path.abspath(conf_file_path), "r", encoding="utf-8") as file:
+        # Read the configuration file - `repos.conf`
+        for line in file:
+            if line.startswith("#"):
+                continue
+            parts = line.split()
+            local_repo = parts[0]
+            layer_name = local_repo.split("/")[-1]
+            token = os.environ.get("GH_PAT")
+            if token is None:
+                fork_url = f"https://github.com/{username}/{layer_name}.git"
+            else:
+                fork_url = (
+                    f"https://x-access-token:{token}@github.com/"
+                    f"{username}/{layer_name}.git"
+                )
+            git_obj = GitRepo(
+                local_repo=local_repo,
+                upstream_repo_url=parts[1],
+                upstream_branch=parts[2],
+                local_base_branch=parts[3],
+                upstream_repo_name=upstream_repo_name,
+                fork_name=fork_name,
+                fork_url=fork_url,
+            )
+            print(f"\n{layer_name}\n")
+            if skip_merge:
+                print("\tUpstream Merge has Been Skipped")
+                merge_report[git_obj] = (0, "Upstream Merge has Been Skipped")
+            else:
+                os.chdir(git_obj.local_repo)
+                merge_report[git_obj] = merge_upstream(
+                    git_obj, force_checkout, merge_branch_name
+                )
+                os.chdir("../..")
+
+    return merge_report
+
+
+def build_and_test(clean_build, rt_target_IP, build_args):
+    """
+    Build images and run tests on a VM if there are no merge errors.
+
+    :param clean_build: Boolean indicating whether to perform a clean build.
+    :param rt_target_IP: The target IP address of the RT system.
+    :param build_args: To build with NI specific arguments.
+    :return: A tuple (status_code, message).
+            Returns (0, None) on success, or error details on failure.
+    """
+    success = setup_env_and_build_packages(build_args, clean_build)
+    if success[0] != 0:
+        return success
+    success = install_and_test_image(rt_target_IP)
+    return success
+
+
+def push_submodules_and_create_PRs(
+    merge_report,
+    merge_branch_name,
+    pr_title,
+    pr_description,
+    username
+):
+    """
+    For each sub-module that was successfully merged and built, push the merge
+    branch to the user's fork and create a pull request targeting the base
+    branch. Updates the push_and_pr_results dictionary with the results of
+    the push and PR steps.
+
+    :param merge_report: Dictionary mapping GitRepo objects to
+                        (status, message).
+    :param merge_branch_name: Name of the branch to push and create PRs from.
+    :param pr_title: Title for the pull request.
+    :param pr_description: Description for the pull request.
+    :param username: GitHub username owning the downstream fork.
+    :return: Dictionary with push and PR results for each sub-module.
+    """
+    push_and_pr_results = {}
+    for git_obj, (status, message) in list(merge_report.items()):
+        if status == 0 and message is not None:
+            os.chdir(git_obj.local_repo)
+            push_and_pr_results[git_obj] = push_branch_and_create_pr(
+                git_obj,
+                merge_branch_name,
+                username,
+                pr_title,
+                pr_description
+            )
+
+    return push_and_pr_results
+
+
+def get_pr_description(work_item_id):
+    """
+    Generate and return the pull request description text.
+
+    :param work_item_id: The work item ID to reference in the PR description.
+    :return: A formatted string containing the PR checklist and work item.
+    """
+    checklist = (
+        "- [x] bitbake packagefeed-ni-core\n"
+        "- [x] bitbake packagegroup-ni-desirable\n"
+        "- [x] bitbake package-index && bitbake nilrt-base-system-image\n"
+        "- [x] Installed BSI on a VM and verified it boots successfully"
+    )
+    work_item_line = (
+        f"\n\nAB#{work_item_id}\n"
+        if work_item_id is not None else ""
+    )
+    return (
+        f"Merge latest from upstream. No conflicts."
+        f"{work_item_line}\n\n{checklist}"
+    )
+
+
+def pull_from_base_branch(branch, upstream_url):
+    """
+    Pull the latest changes from the NILRT repository.
+
+    :param branch: The branch name to pull.
+    :param upstream_url: The URL of the upstream repository.
+    :return: A tuple (status_code, message). Returns (0, None) on success.
+    """
+    git_obj = GitRepo(
+        local_base_branch=branch,
+        upstream_repo_name="upstream",
+        upstream_repo_url=upstream_url,
+    )
+
+    add_remote_details = git_obj.add_remote()
+    if add_remote_details[0] != 0:
+        print(add_remote_details[1])
+        return add_remote_details
+
+    pull_details = git_obj.pull()
+    if pull_details[0] != 0:
+        print(pull_details[1])
+        return pull_details
+
+    return (0, None)
+
+
+def update_meta_nilrt_branch(meta_nilrt_branch):
+    """
+    Pull the latest changes from the meta-nilrt repository.
+
+    :param meta_nilrt_branch: The branch name to pull from the meta-nilrt repo.
+    :return: A tuple (status_code, message). Returns (0, None) on success.
+    """
+    os.chdir("sources/meta-nilrt")
+    # Ensure that the meta-nilrt branch is up to date
+    pull_from_meta_nilrt_details = pull_from_base_branch(
+        meta_nilrt_branch, META_NILRT_URL
+    )
+    os.chdir("../..")
+    if pull_from_meta_nilrt_details[0] != 0:
+        print(pull_from_meta_nilrt_details[1])
+        return pull_from_meta_nilrt_details
+
+    return (0, None)
+
+
+def main():
+    """
+    Main entry point for the upstream merge automation script.
+    Parses arguments, updates repositories, performs merges, builds, tests,
+    pushes, creates PRs, and sends a summary email.
+    """
+    args = parse_args()
+    skip_merge = args.s
+
+    json_config_obj = JsonConfig(
+        automation_conf_path=args.c,
+        work_item_id=args.w
+    )
+
+    setup_logging(json_config_obj.log_level)
+
+    # Ensure that the NILRT branch is up to date (important if repos.conf is
+    # modified)
+    pull_from_nilrt_details = pull_from_base_branch(
+        json_config_obj.nilrt_branch,
+        NILRT_URL
+    )
+    if pull_from_nilrt_details[0] != 0:
+        print(pull_from_nilrt_details[1])
+        return
+
+    merge_report = merge_submodules_with_upstream(
+        json_config_obj.conf_file_path,
+        json_config_obj.force_checkout,
+        json_config_obj.username,
+        json_config_obj.upstream_repo_name,
+        json_config_obj.merge_branch_name,
+        json_config_obj.fork_name,
+        skip_merge,
+    )
+
+    merge_has_errors = any(
+        status != 0 for status, _ in merge_report.values()
+    )
+
+    update_meta_nilrt_branch(json_config_obj.meta_nilrt_branch)
+
+    if merge_has_errors:
+        print(
+            "Merge has errors, skipping build and test."
+        )
+        build_and_test_details = (1, "Merge has Errors")
+    else:
+        build_and_test_details = build_and_test(
+            clean_build=skip_merge,
+            rt_target_IP=json_config_obj.rt_target_IP,
+            build_args=json_config_obj.build_args,
+        )
+
+        if build_and_test_details[0] == 0 and not args.skip_push_and_pr:
+            push_and_pr_results = push_submodules_and_create_PRs(
+                merge_report,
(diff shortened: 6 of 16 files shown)


.github/workflows/manual.yml
```diff
@@ -0,0 +1,54 @@
...
+        type: string
+
+      workItemID:
+        description: 'Provide Work Item ID for the upstream merge script'
+        default: 'None'
+        required: true
+        type: string
+
+      skip_merge:
+        description: 'Skip the merge step'
```
pratheekshasn: 'Skip the merge step.'

Can you explain the use-case in the description? For example, that this is to be used when there's a merge conflict and one of us logs into the dedicated machine, fixes the conflicts and commits them. And that after that, this option of the pipeline can be used.

pratheekshasn: If you haven't specified how to do this in the README, could you add that as well? Basically describe both the ways this can be handled:
1. Log into the machine, fix conflicts, and directly run the python script to push the branches and create the PR.
2. Log into the machine, fix conflicts, and run the pipeline with the skip_merge option

.github/workflows/manual.yml
```diff
@@ -0,0 +1,54 @@
...
+        env:
+          GH_PAT: ${{ secrets.GH_PAT }}
+        run: |
+          if [ "${{ inputs.skip_merge }}" = "True" ]; then
+            echo "Skipping checkout, using pre-cloned repo"
+            cd /home/tonks/dev/CommitSentience/actions-runner/_work/nilrt/nilrt
+          fi
+
+          echo "Running script: ${{ inputs.filename }}"
+          python3 "${{ inputs.filename }}" -w "${{ inputs.workItemID }}" -s "${{ inputs.skip_merge }}"
```
pratheekshasn: You might want to add a trailing new line at the end of the file.

.github/workflows/manual.yml
```diff
@@ -0,0 +1,54 @@
+name: Manual workflow
+
+on:
+  workflow_dispatch:
+    inputs:
+      filename:
+        description: 'Provide the path to the upstream merge script'
+        default: 'scripts/dev/upstream_merge/upstream_merge_and_test.py'
```
pratheekshasn: default: 'scripts/dev/upstream_merge/upstream_merge_and_test.py'

Should this be an input at all? We know this is the file we want to run every time, right? Should it be user-configurable at all?

.github/workflows/manual.yml
```diff
@@ -0,0 +1,54 @@
...
+        required: true
+        type: string
+
+      workItemID:
+        description: 'Provide Work Item ID for the upstream merge script'
```
pratheekshasn: description: 'Provide Work Item ID for the upstream merge script. This is the work item to which the PRs created will link to`.

Or rephrase it to something similar.

scripts/dev/upstream_merge/GitRepo.py
```diff
@@ -0,0 +1,93 @@
+from git_commands import *
+
+class GitRepo:
+    def __init__(self, local_repo=None, upstream_repo_url=None, upstream_branch=None, local_base_branch=None,upstream_repo_name=None,fork_name=None,fork_url=None):
+        self.local_repo = local_repo
```
pratheekshasn: `self.local_repo = local_repo`

Is this the name or path? You might want to change the name to denote that as well.

scripts/dev/upstream_merge/GitRepo.py
```diff
@@ -0,0 +1,93 @@
...
+        self.fork_name = fork_name
+        self.fork_url = fork_url
+
+    def get_current_commit(self):
+        """
+        Get the current HEAD commit hash.
+        :param capture_output: Whether to capture the command output.
+        :return: The current commit hash.
+        """
+        return run_command("git rev-parse HEAD")[1]
```
pratheekshasn: `git rev-parse HEAD`

Do you think it makes sense to shift this to git_commands.py and just call that here instead?
@rajendra-desai-ni

scripts/dev/upstream_merge/GitRepo.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        Check if a branch exists locally.
+        :param branch_name: Name of the branch to check.
+        :param capture_output: Whether to capture the command output.
+        :return: True if the branch exists, False otherwise.
+        """
+        if check_on_remote:
+            if fork_name is None:
+                fork_name = self.upstream_repo_name
+            result = run_command(f"git ls-remote --heads {fork_name} {branch_name}", capture_output)
```
pratheekshasn: `git ls-remote --heads {fork_name} {branch_name}`

Do you think it makes sense to shift this to git_commands.py and just call that here instead?
@rajendra-desai-ni

pratheekshasn:
Could you please run pylint through all the files? And files like build, test, shell_commands, git_commands aren't part of upstream_merge specifically, these should be usable by anyone. Could you move the appropriate file to outside the upstream_merge folder?

You also need to add more details in the PR description about what files you're adding in this PR and why.

scripts/dev/upstream_merge/GitRepo.py
```diff
@@ -0,0 +1,93 @@
...
+    
+    def checkout_branch(self, branch_name, create = False, force_checkout = False):
+        """Switch to the given branch."""
+        return git_checkout(branch_name, create, force_checkout)
+
+    def delete_branch(self,branch_name):
+        """Delete a local branch."""
+        return git_branch(branch_name,delete=True)
+
+    def fetch_branch(self, upstream_repo_name = None, repo_branch = None):
```
pratheekshasn: `def fetch_branch(self, upstream_repo_name = None, repo_branch = None):`

Can the above be changed to this instead? Would it still result in the same meaning?

`def fetch_branch(self, upstream_repo_name = self.upstream_repo_name, repo_branch = self.upstream_branch):`

.github/workflows/manual.yml
pratheekshasn: Here, and the all the other files as well.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,202 @@
+# MOTIVATION
+
+To ensure the long-term maintainability and efficiency of the ni/nilrt repository, automating the upstream merge process is a crucial step. Manual upstream merges are often tedious, time-consuming, and prone to human error, especially as the size and frequency of upstream changes grow. By automating this process, we can significantly reduce the cost of maintenance by saving several hours of manual effort per release cycle. This allows developers to focus more on value-adding tasks rather than routine integration work. Automation also ensures that our fork stays closely aligned with upstream changes, minimizing future integration conflicts and making it easier to adopt new features, security patches, and bug fixes promptly. Overall, this improves the quality, stability, and security of the codebase while enhancing the team's productivity and responsiveness to upstream evolution.
+
+---
+
```
pratheekshasn: Add another section here describing what all the script accomplishes. Mention that it not only does an upstream merge, but also takes care of building and testing all the new changes, pushing the merged commits and creating pull requests to the ni repos as well.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,202 @@
...
+### **2. Run the Script**
+#### **To Perform a Merge**
+This will merge upstream changes, build the images, and test them:
+```bash
+python3 scripts/dev/upstream_merge/upstream_merge.py -w workItemID
+```
+
+**`Note`**:
+- If the configuration file is `automation_conf.json`, you do not need to specify its path explicitly, as it is set as default. However, if you are using a different configuration file, you must provide its path using the `-c` argument.For example:
```
pratheekshasn: You can move this automation_conf.json note to after you have explained what that JSON file is, otherwise it doesn't make sense to bring it up before you define it.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+```bash
+python3 scripts/dev/upstream_merge/upstream_merge.py -c path/to/your_config.json -w workItemID
+```
+
+- **`workItemID`**:  
+The `workItemID` associated with the pull request. If `workItemID` is not provided `None` will be used.
+
+#### **To Skip the Merge and Only Build and Test**
+If you want to skip the merge step and directly proceed with building and testing:
```
pratheekshasn: >If you want to skip the merge step and directly proceed with building and testing:

Why would I want to do this? You'll need to explain this too.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,202 @@
...
+```bash
+python3 scripts/dev/upstream_merge/upstream_merge.py -w workItemID -s True
+```
+
+---
+
+## **Configuration File: `automation_conf.json`**
+
+The script relies on a configuration file (`automation_conf.json`) to define various parameters. **All fields are required**, and if a field is not present, `None` will be used as the default value. Below is an explanation of the fields in the configuration file:
```
pratheekshasn: Instead of saying the script relies on a particular file, say that the script is user configurable. That sounds like it will sell better :)

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,202 @@
...
+ 
+- **`NILRT_branch`**:
+  The nilrt branch to pull the latest changes from.
+
+- **`meta_nilrt_branch`**:
+  The meta-nilrt branch to pull the latest changes from.
```
pratheekshasn, line 52: Remind me why we need the meta-nilrt branch again? We definitely don't pull the latest changes from it, because that's a layer that NI has built, we don't pull it from any upstream.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,202 @@
...
+     - **Runmode Image**
+
+5. **Install Images on the RT Target (VM)**:  
+   - The safemode and runmode images are installed on a Real-Time (RT) target, which is a Virtual Machine in this case.
+
+6. **Test the Installation**:  
+   - The script verifies that the installation is not faulty by running tests on the VM.
+
+**Important Note**
+Currently, the VM and the snapshot required for this script exist only on the tonks machine. As a result, the script can only be run on this machine. In the future, we plan to set up a YAML-based script (or a similar automation tool) to run the entire process on this machine.
```
pratheekshasn: Let's not mention the name of the machine.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,202 @@
...
+- Acts as a wrapper for running system commands from Python.
+
+### **`git_commands.py`**
+- Contains various Git-related functions that rely on `shell_commands.py`.  
+- Includes operations like fetching, pulling, creating branches, and merging.
+
+### **`GitRepo.py`**
+- Defines the `GitRepo` class, which encapsulates details and operations for a Git repository.  
+-  Data members are:
+    - local_repo
```
pratheekshasn, line 172: `local_repo`

Could you mention/rename according to whether this is the name or the path?

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,202 @@
...
+- Handles the process of building images.  
+- Key steps include:
+  - Setting up the Docker environment.
+  - Building core feeds.
+  - Building core images.
+
+### **`test.py`**
+- Manages testing of the built images on a VM.  
+- Key steps include:
+  - Restoring the VM to a specific snapshot.
```
pratheekshasn: Mention that it's restored to a clean point with SSH enabled.

scripts/dev/upstream_merge/build.py
```diff
@@ -0,0 +1,82 @@
...
+        return docker
+    print("\nDocker setup completed.")
+
+    if clean_build:
+        # Step 2: Clean the build feeds and images
+        clean = clean_build_feeds_and_images(org)
+        if clean[0] != 0:
+            return clean
+        print("\nClean build feeds and images completed.")
+    else:
```
pratheekshasn: Should there be an `else` block here? If I do specify clean_build, it will only clean and not actually build.

scripts/dev/upstream_merge/build.py
```diff
@@ -0,0 +1,82 @@
...
+        # Step 3: Build the core images
+        core_images = build_core_images(org)
+        if core_images[0] != 0:
+            return core_images
+        print("\nCore images build completed.")
+
+    # Return success if all steps are completed
+    return (0, None)
+
+def clean_build_feeds_and_images(org=""):
```
pratheekshasn: Could you rename the parameter to `args` instead of `org`?

scripts/dev/upstream_merge/build.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Build the core images.
+    This step involves running the script to build the core images for the project.
+    :return: A tuple (status_code, message). Returns (0, None) on success.
+    """
+    print("\nBuilding core images...\n")
+    return execute_and_stream_cmd_output(f"bash scripts/pipelines/build.core-images.sh {org}")
+
+if __name__ == "__main__":
+    status_code, message = build_images()
```
pratheekshasn: If I were to directly run this module with the arg `--org`, how would I do it? Don't you need to pass the arguments here in the main function as well to build_images()?

scripts/dev/upstream_merge/git_commands.py
```diff
@@ -0,0 +1,176 @@
...
+    :param directory: Directory to clone into (optional).
+    :param depth: Depth for shallow cloning (optional).
+    :param capture_output: Whether to capture the command output.
+    """
+    command = f"git clone {repo} {directory}" if directory else f"git clone {repo}"
+    if depth:
+        command += f" --depth {depth}"
+    return run_command(command, capture_output)
+
+def git_commit(message, amend=False, capture_output=True):
```
pratheekshasn: Could you add the --signoff option too in git_commit()?

scripts/dev/upstream_merge/push_and_PR.py
```diff
@@ -0,0 +1,66 @@
...
+        return PR_details
+    
+    return (0,None)
+
+def get_PR_description(work_item_id):
+    return f"""Merge latest from upstream. No conflicts.
+ 
+    #AB{work_item_id}
+    
+    - [ ] bitbake packagefeed-ni-core
```
pratheekshasn: - [  ] will need to be changed to - [x]

Reimaged a **NILRT VM** with the new base image and successfully booted it

scripts/dev/upstream_merge/push_and_PR.py
pratheekshasn, line 1: Isn't there a main() for this module?

scripts/dev/upstream_merge/test.py
```diff
@@ -0,0 +1,161 @@
+from shell_commands import *
+import time
+import os
+
+def OS_test(VM_name, snapshot_name):
```
pratheekshasn: Could you rename this to install_and_test_image()?

scripts/dev/upstream_merge/test.py
```diff
@@ -0,0 +1,161 @@
...
+    Steps:
+    1. Restore the VM to a specific snapshot.
+    2. Start the VM.
+    3. Copy, extract and install the safemode image.
+    4. Copy, extract and install the runmode image.
+    5. Verify the OS version.
+    6. Power off the VM.
+    """
+    username="admin"
+    target="NI-cRIO-903x-VM-27108694"
```
pratheekshasn: Would it make more sense to put the target name, VM name, VM check point name on a file on disk instead of specifying it on the repo? We could read these values from a file and the file path can be set as an environment variable on the host machine as part of setting up the host machine. What are your thoughts? @rajendra-desai-ni, @prasanna-nib

scripts/dev/upstream_merge/test.py
```diff
@@ -0,0 +1,161 @@
...
+    print("Starting the VM...")
+    return execute_and_stream_cmd_output(f"VBoxManage startvm \"{VM_name}\" --type headless")
+
+def copy_image(filename ,ssh_connection):
+    """
+    Copy the image to the target machine.
+    """
+    print("Copying image to target machine...")
+    current_directory = os.getcwd()
+    return execute_and_stream_cmd_output(f"scp {current_directory}/build/tmp-glibc/deploy/images/x64/{filename} {ssh_connection}:/home/admin")
```
pratheekshasn: Could you make this path `/build/tmp-glibc/deploy/images/x64/` a constant specified on the top of this file somewhere? Same for the names of the safemode and runmode tar files. That way you don't have to copy paste the names in more than one place, you can just use the constant.

scripts/dev/upstream_merge/upstream_merge.py
```diff
@@ -0,0 +1,119 @@
+import os
+from GitRepo import *
+
+def merge_submodules_with_upstream(conf_file, force_checkout, username, upstream_repo_name, merge_branch_name, fork_name, skip_merge):
+    merge_report = {}
+    current_directory = os.getcwd()
+
+    with open(conf_file, "r") as file: # Read the configuration file - `repos.conf` and not the `automation_conf.json`
```
pratheekshasn: No need to specify "and not the automation_conf.json", you can just say it's for repos.conf.

scripts/dev/upstream_merge/upstream_merge.py
```diff
@@ -0,0 +1,119 @@
...
+                            fork_name=fork_name,
+                            fork_url=fork_url)
+            os.chdir(git_obj.local_repo)
+            merge_report[git_obj] = merge_upstream(git_obj, force_checkout, merge_branch_name, skip_merge)
+            os.chdir(current_directory)
+    return merge_report
+
+def merge_upstream(git_obj,force_checkout, merge_branch_name, skip_merge):
+    if skip_merge == "True":
+        return (0," Has Been Skipped")
```
pratheekshasn: What has been skipped? You might want to write a clearer message.

scripts/dev/upstream_merge/GitRepo.py
```diff
@@ -0,0 +1,93 @@
...
+        if upstream_repo_name in existing_remotes:
+            git_remote(upstream_repo_name,remove=True)
+
+        return git_remote(upstream_repo_name, repo_url)
+
+    def diff(self):
+        """Check if there are differences in the last merge."""
+        return git_diff(compare_with="HEAD~1", capture_output=True)
+
+    def pull_latest(self, branch_name = None, upstream_repo_name = None):
```
pratheekshasn: Just `pull()` should suffice as the name, it's usually always the latest.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,186 @@
...
+    os.makedirs(f"temp/{datetime.datetime.now().strftime('%d-%m-%Y')}", exist_ok=True)
+    logging.basicConfig(
+        level=log_level,
+        format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+        handlers=[
+            logging.FileHandler(f"temp/{datetime.datetime.now().strftime('%d-%m-%Y')}/upstream_merge_{datetime.datetime.now().strftime('%H-%M-%S')}.log", mode='a')
+        ]
+    )
+    
+def write_email_addresses(email_log_file_name, email_from, email_to):
```
pratheekshasn: Can be renamed to write_email_headers()

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,186 @@
...
+    sources/bitbake
+     ... ERRORS
+    """
+    min_detail = ""
+    error_detail = ""
+    diff_detail = ""
+
+    build_and_test_detail = merge_report.pop("Build and Test")
+    if build_and_test_detail[0] == 0:
+        push_and_PR_details = merge_report.pop("Push and PR")
```
pratheekshasn: You should add a comment here and explain why you're doing a pop.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,186 @@
...
+    else:
+        min_detail += " ... ERRORS\n"
+        error_detail += f"\nBuild and Test\n ... ERRORS\n    {build_and_test_detail[1]}\n"
+        diff_detail += f"\nBuild and Test\n ... ERRORS\n    {build_and_test_detail[1]}\n"
+    
+    if email_log_level == 0:
+        return min_detail + "\n\n" + error_detail
+    return min_detail + "\n\n" + error_detail + "\n\n" + diff_detail
+
+def write_log(email_log_file_name, contents):
```
pratheekshasn: Isn't this just the log_file_name? And not email_log_file_name?

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,186 @@
...
+
+def main():
+    args = parse_args()
+    skip_merge = args.s
+
+    json_config_obj = json_config(automation_conf_path=args.c,workitemID=args.w)    
+    
+    setup_logging(json_config_obj.log_level)
+
+    pull_from_nilrt_details = pull_from_base_branch(json_config_obj.NILRT_branch,"https://github.com/ni/nilrt.git") # To ensure that the NILRT branch is up to date in case files like 'repos.conf' are modified, which would be crucial to the current script
```
pratheekshasn: Could you declare this URL ("https://github.com/ni/nilrt.git") as a constant somewhere in the file and use the constant here instead?

pratheekshasn (changes requested):

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,202 @@
...
+### **1. Navigate to the root of the nilrt repo**
+```bash
+cd ~/nilrt
+```
+
+### **2. Run the Script**
+#### **To Perform a Merge**
+This will merge upstream changes, build the images, and test them:
+```bash
+python3 scripts/dev/upstream_merge/upstream_merge.py -w workItemID
```
rajendra-desai-ni: I hope [upstream_merge_and_test.py](https://github.com/ni/nilrt/blob/949f7e451b0432c8b6d0e20dc1fc688c93aa3a37/scripts/dev/upstream_merge/upstream_merge_and_test.py#L1) is the script you are referring to here.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+**`Note`**:
+- If the configuration file is `automation_conf.json`, you do not need to specify its path explicitly, as it is set as default. However, if you are using a different configuration file, you must provide its path using the `-c` argument.For example:
+
+```bash
+python3 scripts/dev/upstream_merge/upstream_merge.py -c path/to/your_config.json -w workItemID
```
rajendra-desai-ni: I hope [upstream_merge_and_test.py](https://github.com/ni/nilrt/blob/949f7e451b0432c8b6d0e20dc1fc688c93aa3a37/scripts/dev/upstream_merge/upstream_merge_and_test.py#L1) is the script you are referring to here.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+- **`workItemID`**:  
+The `workItemID` associated with the pull request. If `workItemID` is not provided `None` will be used.
```
rajendra-desai-ni: I guess workItemID would show up in the PRs? Can we just skip workItemID if it's not provided, instead of setting it to 'None' since we are trying to use a pipeline and we would not be having a workItemID until a work item is created.

rajendra-desai-ni (changes requested):

scripts/dev/utils/build.py
```diff
@@ -0,0 +1,90 @@
+"""Functions for building images and related tasks in the upstream merge workflow."""
+
+from shell_commands import execute_and_stream_cmd_output
+
+def build_images(org="", clean_build=False):
```
rajendra-desai-ni: Could you rename this function to be a bit more specific? We are not just building images, but cleaning feeds, building container, building feeds and then images right? Also, there is already a similar function name that just builds images (build_core_images), so it should not be confusing with this one.

scripts/dev/utils/push_and_pr.py
```diff
@@ -0,0 +1,118 @@
+"""Functions for pushing branches to forks and creating pull requests as part of the upstream merge\
+    workflow."""
+
+import os
+
+def push_and_pr_prepare(merge_has_errors, build_and_test_details, merge_report,
```
rajendra-desai-ni: This function and below function (push_branch_and_create_pr) seems to be confusing. Could you change either of these 2 so that its less confusing? Also, the function descriptions.

scripts/dev/utils/test.py
```diff
@@ -0,0 +1,176 @@
...
+        return os_version
+
+    # Step 6: Power off the vm
+    poweroff = poweroff_vm(vm_name)
+    if poweroff[0] != 0:
+        return poweroff
+
+    return os_version
+
+def install_safemode_image(ssh_connection):
```
rajendra-desai-ni: This function doesn't actually install safemode image but `extract_and_install_safemode_image` function does that. Function name can be misleading, you can either combine the 2 functions into 1 or rename this function.

scripts/dev/utils/test.py
```diff
@@ -0,0 +1,176 @@
...
+
+    reboot = reboot_machine(ssh_connection)
+    if reboot[0] != 0:
+        return reboot
+
+    time.sleep(90)  # Wait for the machine to reboot
+
+    return (0, None)
+
+def install_runmode_image(ssh_connection):
```
rajendra-desai-ni: This function doesn't actually install runmode image but `extract_and_install_runmode_image` function does that. Function name can be misleading, you can either combine the 2 functions into 1 or rename this function.

scripts/pipelines/clean_build.core-feeds_and_core-images.sh
```diff
@@ -0,0 +1,21 @@
...
+
+SCRIPT_ROOT=$(realpath $(dirname "${BASH_SOURCE[0]}"))
+
+SCRIPT_ROOT1=${SCRIPT_ROOT}
+
+. "${SCRIPT_ROOT}/build.common.sh" $@
+
+echo "INFO: Cleaning and building the core package feed."
+
+bitbake -c cleanall packagefeed-ni-core
```
rajendra-desai-ni, line 9: Why do we need to do cleanall before building? Won't we be doing a fresh build each time?

rajendra-desai-ni (changes requested):

scripts/dev/upstream_merge/GitRepo.py
CommitSentience: It is the name on repos.conf

.github/workflows/manual.yml
CommitSentience: Has been fixed.

.github/workflows/manual.yml
CommitSentience: Have looked into it.

.github/workflows/manual.yml
CommitSentience: Has been fixed.

.github/workflows/manual.yml
CommitSentience: It shouldn't be an input. I have removed it from taking an input.

scripts/dev/upstream_merge/README.md
CommitSentience: I have added another paragraph which talks about its accomplishments.

scripts/dev/upstream_merge/README.md
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/README.md
CommitSentience: I have explained it.

scripts/dev/upstream_merge/README.md
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/README.md
CommitSentience: This is done to ensure that the meta-nilrt layer is upto date. The nilrt repo generally points to older hash commits.
I had faced an issue with this when meta-nilrt wasn't upto to date and it was failing at dkms as it was trying to pull from a branch which didn't exist anymore. The newer commits had the required changes.

scripts/dev/upstream_merge/README.md
CommitSentience: It is the name on repos.conf

scripts/dev/upstream_merge/README.md
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/build.py
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/build.py
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/build.py
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/git_commands.py
CommitSentience: I have added it.

scripts/dev/upstream_merge/test.py
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/upstream_merge_and_test.py
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/GitRepo.py
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/upstream_merge.py
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/upstream_merge_and_test.py
CommitSentience: I have added it.

scripts/dev/upstream_merge/README.md
CommitSentience: Yes, It is.
It has been updated.

scripts/dev/upstream_merge/README.md
CommitSentience: Yes, It is.
It has been updated.

scripts/dev/upstream_merge/README.md
pratheekshasn: Could you explain this in the code comments please?

scripts/dev/upstream_merge/GitRepo.py
rajendra-desai-ni: Yes, I think its better to move this to git_commands.py that already implements other git specific tasks. I am also fine with existing change if it takes lot of effort to move.

scripts/dev/upstream_merge/GitRepo.py
rajendra-desai-ni: Yes, I think its better to move this to git_commands.py that already implements other git specific tasks. I am also fine with existing change if it takes lot of effort to move.

.github/workflows/manual.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: git config pull.rebase false
+
+      - name: Run Python upstream merge script
+        env:
+          GH_PAT: ${{ secrets.GH_PAT }}
+        run: |
+          if [ "${{ inputs.skip_merge }}" = "True" ]; then
+            username=$(jq -r '.username' scripts/dev/upstream_merge/automation_conf.json)
+            echo "Skipping checkout, using pre-cloned repo"
+            cd /home/tonks/dev/${username}/actions-runner/_work/nilrt/nilrt
```
pratheekshasn: I don't think this should be set to `/home/tonks/...`. If someone else clones this repo and tries to run the pipeline, it will fail with a path error. The runner is always listening on the host machine, right? Can it not get the current directory because it's already running on this path?

If that fails, maybe you could think of setting the path to ~/...

scripts/dev/upstream_merge/README.md
pratheekshasn: You left out the most important part. The part where it creates PRs.

The script does the following tasks:
1. Automatically merges changes from upstream repositories into the local fork, reducing manual effort.
1. Builds the safemode and runmode images after a successful merge to ensure the codebase is functional.
1. Installs and tests these images on a virtual machine to verify that the integration is successful and the target works as expected.
1. Commits and pushes the new changes and creates a PR targeting the main branch.
1. Sends email reports about the merge and build status.


At a high level, this is what we want to convey.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,216 @@
...
+## **Steps to Run the Script**
+
+### **1. Navigate to the root of the nilrt repo**
+```bash
+cd ~/nilrt
+```
+
+### **2. Run the Script**
+#### **To Perform a Merge**
+This will merge upstream changes, build the images, and test them:
```
pratheekshasn: This will merge upstream changes, build the images, and test them and creates PR with the new changes after the build and test steps succeed.

Note that these steps are serial. The build should succeed for the test to run, and the test should pass for the PRs to be created.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,216 @@
...
+  
+- **`vm_name`**:  
+  The name of the VM where the images will be tested.
+
+- **`snapshot_name`**:  
+  The snapshot of the VM to be restored before testing. This snapshot must have SSH enabled.
+
+
+**`Note`**:
+- If the configuration file is `automation_conf.json`, you do not need to specify its path explicitly, as it is set as default. However, if you are using a different configuration file, you must provide its path using the `-c` argument.For example:
```
pratheekshasn: [Trivial] Could you add a space before `For example:`?

scripts/dev/upstream_merge/test.py
rajendra-desai-ni: As we are planning to use a pipeline, we can pass VM/target details as an argument to this script while running.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,216 @@
...
+### **`test.py`**
+- Manages testing of the built images on a VM.  
+- Key steps include:
+  - Restoring the VM to a clean point with SSH enabled.
+  - Installing and testing safemode and runmode images.
+  - Verifying the OS version.
+
+### **`push_and_pr.py`**
+- Handles pushing the merged changes to the downstream fork repository.
+- Automates the creation of a pull request (PR) on GitHub after a successful merge and build.
```
pratheekshasn: [Trivial] You can remove the brackets for PR and leave "pull request" as is.

scripts/dev/upstream_merge/README.md
```diff
@@ -0,0 +1,216 @@
...
+     - **Runmode Image**
+
+5. **Install Images on the RT Target (VM)**:  
+   - The safemode and runmode images are installed on a Real-Time (RT) target, which is a Virtual Machine in this case.
+
+6. **Test the Installation**:  
+   - The script verifies that the installation is not faulty by running tests on the VM.
+
+**Important Note**
+Currently, the VM and the snapshot required for this script exist only on one of our local machine. As a result, the script can only be run on this machine. In the future, we plan to set up a YAML-based script (or a similar automation tool) to run the entire process on this machine.
```
pratheekshasn: [Trivial] *... one of our machine**s**.

scripts/dev/upstream_merge/upstream_merge.py
```diff
@@ -0,0 +1,201 @@
...
+import sys
+# Ensure imports work regardless of current working directory
+PROJECT_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../..'))
+os.chdir(PROJECT_ROOT)
+if PROJECT_ROOT not in sys.path:
+    sys.path.insert(0, PROJECT_ROOT)
+
+from scripts.dev.utils.git_repo import GitRepo
+
+def merge_submodules_with_upstream(
```
pratheekshasn: I think it makes more sense to move this particular function to upstream_merge_and_test.py. It doesn't make sense as an exported function, the module "upstream_merge" shouldn't know about submodules or repos.conf.

scripts/dev/upstream_merge/upstream_merge.py
```diff
@@ -0,0 +1,201 @@
...
+
+    if merge_result[0] == 0:
+        diff_output = git_obj.diff()
+        if (git_obj.get_current_commit() == commit_before_merge) or diff_output == (0, ''):
+            return (0, None)
+        return (0, diff_output[1])
+
+    return (1, merge_result[1])
+
+def prepare_for_merge(git_obj, merge_branch_name, force_checkout):
```
pratheekshasn: Could you refactor this a bit? Make it handle only these two things:
1. switch to base branch
2. fetch upstream

Create the merge branch in the caller. Let's not have the same function handle these disconnected things.

scripts/dev/upstream_merge/json_config.py
```diff
@@ -0,0 +1,32 @@
...
+        try:
+            with open(automation_conf_path, "r", encoding="utf-8") as file:
+                config = json.load(file)
+        except (FileNotFoundError, json.JSONDecodeError) as err:
+            print(f"Error loading config file: {err}")
+            sys.exit(1)
+        self.work_item_id = work_item_id
+        self.nilrt_branch = config.get("nilrt_branch")
+        self.meta_nilrt_branch = config.get("meta_nilrt_branch")
+        self.conf_file = os.getcwd()+f"/{config.get('conf_file_path')}"
```
pratheekshasn: It makes more sense to name this conf_file_path, since it's the path and not the file descriptor.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,346 @@
...
+    """
+    if merge_has_errors:
+        return (1, "Merge has Errors")
+    success = setup_env_and_build_packages("--org", clean_build)
+    if success[0] != 0:
+        return success
+    success = install_and_test_image(vm_name, snapshot_name)
+    return success
+
+def push_and_pr_prepare(merge_has_errors, build_and_test_details, merge_report,
```
pratheekshasn: You can rename this to push_submodules_and_create_PRs()

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,346 @@
...
+        json_config_obj.merge_branch_name,
+        json_config_obj.fork_name,
+        skip_merge
+    )
+
+    merge_has_errors = any(status != 0 for status, _ in merge_report.values())
+
+    update_meta_nilrt_branch(json_config_obj.meta_nilrt_branch)
+
+    build_and_test_details = build_and_test(
```
pratheekshasn: You can check right here if merge has errors, before calling subsequent functions.

if not merge_has_errors:
    build_and_test
    if not build_and_test_details.status: # or whatever that variable is called
        push_and_create_PR()

write_email

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,346 @@
...
+    :param email_from: The sender's email address.
+    :param email_to: The recipient's email address.
+    :param merge_report: The merge report data structure.
+    :param email_log_level: The log verbosity level for the email.
+    """
+    email_file_name = (
+        f"temp/{datetime.datetime.now().strftime('%d-%m-%Y')}/"
+        f"upstream_merge_{datetime.datetime.now().strftime('%H-%M-%S')}.txt"
+    )
+    formatted_report_string = format_merge_report(merge_report, email_log_level)
```
pratheekshasn: Could you move all these formatting and log related functions to a new file?

scripts/dev/utils/build.py
```diff
@@ -0,0 +1,123 @@
...
+from scripts.dev.utils.shell_commands import execute_and_stream_cmd_output
+
+def setup_env_and_build_packages(build_args="", clean_build=False):
+    """
+    Build the images for the project.
+    This function orchestrates the steps required to build the images, including
+    setting up the Docker environment, cleaning build feeds and images, and building
+    the core feeds and images.
+
+    :param build_args: (Optional) String of extra arguments to pass to the build scripts.
```
pratheekshasn: --org isn't really a build argument, it's used for setting up the environment.

scripts/dev/utils/build.py
```diff
@@ -0,0 +1,123 @@
...
+    print("\nCore feeds build completed.")
+
+    # Step 3: Build the core images
+    core_images = build_core_images(build_args)
+    if core_images[0] != 0:
+        return core_images
+    print("\nCore images build completed.")
+
+    # Return success if all steps are completed
+    return (0, None)
```
pratheekshasn: You can return (0, "Building core feeds, safemode and runmode succeeded")

scripts/dev/utils/git_commands.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    command = "git commit"
+    if amend:
+        command += " --amend"
+    if signoff:
+        command += " --signoff"
+    command += f' -m "{message}"'
+    return run_command(command, capture_output)
+
+def git_push(remote_repo_name="origin", branch="main", force=False, delete=False,
```
pratheekshasn: Could you please rename the parameter name to deleteExisting?

scripts/dev/utils/git_commands.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    if name is not None:
+        if url is not None:
+            return run_command(f"git remote add {name} {url}", capture_output)
+        if remove:
+            return run_command(f"git remote remove {name}", capture_output)
+
+    return run_command("git remote", capture_output)
+
+def git_merge(branch_name, message="Merge latest upstream", no_ff=False, signoff=False,
```
pratheekshasn: This module shouldn't care about merging from upstream, it's a generic module that can be reused by anyone. Let's not have the default message set to this.

scripts/dev/utils/git_repo.py
```diff
@@ -0,0 +1,107 @@
...
+from scripts.dev.utils.git_commands import (
+    run_command, git_checkout, git_branch, git_fetch, git_merge, git_remote,
+    git_pull, git_push, git_pull_request, git_diff
+)
+class GitRepo:
+    """Encapsulates details and operations for a Git repository, including remotes, branches,\
+        and merges."""
+    def __init__(self, local_repo=None, upstream_repo_url=None, upstream_branch=None,
+                 local_base_branch=None,upstream_repo_name=None,fork_name=None,fork_url=None):
+        self.local_repo = local_repo # Path to the local repository as mentioned in `repos.conf`.
```
pratheekshasn: Would it be easier to just store the entire absolute path of the repo here instead of the relative path?

scripts/dev/utils/push_and_pr.py
pratheekshasn, line 1: Could you rename this to push_branch_and_create_pr?

scripts/dev/utils/push_and_pr.py
```diff
@@ -0,0 +1,125 @@
...
+        git_obj.local_base_branch,
+        f"{username}:{merge_branch_name}"
+    )
+    if pr_details[0] != 0:
+        print(pr_details[1])
+        return pr_details
+
+    return (0, None)
+
+def get_pr_description(work_item_id):
```
pratheekshasn: If this file is in utils, then it should be usable by everyone, let's not have it know about upstream merge PR description. Could you move this to upstream_merge_and_test and pass the whole PR description here?

scripts/dev/utils/push_and_pr.py
```diff
@@ -0,0 +1,125 @@
...
+def parse_args():
+    """
+    Parse command-line arguments for the Push and PR script.
+
+    :return: Parsed arguments as returned by argparse.ArgumentParser.parse_args().
+    """
+    parser = argparse.ArgumentParser(description="Push and PR script")
+    parser.add_argument("-w", type=str, help="Work item ID", default=None)
+    parser.add_argument("-u", type=str, help="Username", default=None)
+    parser.add_argument("-mbr", type=str, help="Merge branch name", default=None)
```
pratheekshasn: Call this branch_name, calling it merge branch name wouldn't make sense for it as an independent generic function.

We are currently using this function for the sake of upstream merging, that doesn't mean everyone who uses this API needs to know about that.

scripts/dev/utils/push_and_pr.py
```diff
@@ -0,0 +1,125 @@
...
+    - [x] Reimaged a cRIO with the new base image and successfully booted it"""
+
+def parse_args():
+    """
+    Parse command-line arguments for the Push and PR script.
+
+    :return: Parsed arguments as returned by argparse.ArgumentParser.parse_args().
+    """
+    parser = argparse.ArgumentParser(description="Push and PR script")
+    parser.add_argument("-w", type=str, help="Work item ID", default=None)
```
pratheekshasn: If you pass in the whole PR description like I mentioned in my previous comment, you won't need this argument.

scripts/dev/utils/test.py
```diff
@@ -0,0 +1,187 @@
...
+    Reboot the target machine.
+    """
+    print("Rebooting the machine...")
+    return execute_and_stream_cmd_output(f'ssh {SSH_CONNECTION} "reboot"')
+
+def verify_os_version():
+    """
+    Verify the os version on the target machine.
+    """
+    print("Verifying os version...")
```
pratheekshasn: [Trivial] Could you capitalize OS? Here and everywhere else?

scripts/pipelines/clean.core-feeds_and_core-images.sh
pratheekshasn, line 1: This file needs an extra newline at the end.

pratheekshasn (changes requested):

scripts/dev/utils/push_and_pr.py
```diff
@@ -0,0 +1,125 @@
...
+    parser.add_argument("-u", type=str, help="Username", default=None)
+    parser.add_argument("-mbr", type=str, help="Merge branch name", default=None)
+    parser.add_argument("-fn", type=str, help="Fork name", default=None)
+    args = parser.parse_args()
+    return args
+
+if __name__ == "__main__":
+    args = parse_args()
+    git_obj = GitRepo()
+    git_obj.fork_name = args.fn if args.fn else git_obj.fork_name
```
pratheekshasn: Isn't the else case here always going to be an empty string?

scripts/pipelines/clean.core-feeds_and_core-images.sh
CommitSentience: I have added a newline at the end.

scripts/dev/utils/test.py
CommitSentience: It has updated.

scripts/dev/utils/push_and_pr.py
CommitSentience: It has been updated.

scripts/dev/utils/push_and_pr.py
CommitSentience: It has been updated.

scripts/dev/utils/push_and_pr.py
CommitSentience: It has been updated.

scripts/dev/utils/git_repo.py
CommitSentience: It has been updated.

scripts/dev/utils/git_commands.py
CommitSentience: It has been updated.

scripts/dev/utils/git_commands.py
CommitSentience: It has been updated.

scripts/dev/utils/build.py
CommitSentience: It has been updated.

scripts/dev/utils/build.py
CommitSentience: I renamed build_args to args.

scripts/dev/upstream_merge/upstream_merge_and_test.py
CommitSentience: It has been done.

scripts/dev/upstream_merge/upstream_merge_and_test.py
CommitSentience: It has been updated.

scripts/dev/upstream_merge/upstream_merge_and_test.py
CommitSentience: It has been updated.

scripts/dev/upstream_merge/json_config.py
CommitSentience: It has been updated.

scripts/dev/upstream_merge/upstream_merge.py
CommitSentience: It has been updated.

scripts/dev/upstream_merge/upstream_merge.py
CommitSentience: It has been updated.

scripts/dev/upstream_merge/README.md
CommitSentience: It has been updated.

scripts/dev/upstream_merge/README.md
CommitSentience: It has been updated.

scripts/dev/upstream_merge/README.md
CommitSentience: It has been updated.

scripts/dev/upstream_merge/README.md
CommitSentience: It has been updated.

.github/workflows/manual.yml
pratheekshasn: image
You can remove the lines I have highlighted in red, since they now seem redundant after you added more clarification after the motivation section.

About the workitemID not being specified, if you have made that change in creating the PR description, please update the README accordingly.

For the highlight in blue, you'll need to specify how they can manually make the changes. They would have to log into the machine on which the merge runs and then do it. You can say that they can contact yourself or me to get information about the credentials of the build machine.

pratheekshasn: image
[Trivial] Extra asterisks in the header.

scripts/dev/upstream_merge/log_and_email_utils.py
```diff
@@ -0,0 +1,163 @@
+
+"""
+Utilities for logging, formatting merge reports, and sending email notifications.
+"""
+
+import os
+import datetime
+import logging
+from scripts.dev.utils.git_commands import send_email
```
pratheekshasn: Did you not have to add these lines to this file?

>PROJECT_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../..'))
os.chdir(PROJECT_ROOT)
if PROJECT_ROOT not in sys.path:
    sys.path.insert(0, PROJECT_ROOT)

scripts/dev/upstream_merge/upstream_merge.py
```diff
@@ -0,0 +1,153 @@
...
+import sys
+# Ensure imports work regardless of current working directory
+PROJECT_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../..'))
+os.chdir(PROJECT_ROOT)
+if PROJECT_ROOT not in sys.path:
+    sys.path.insert(0, PROJECT_ROOT)
+
+from scripts.dev.utils.git_repo import GitRepo
+
+def merge_upstream(git_obj, force_checkout, merge_branch_name, skip_merge):
```
pratheekshasn: This function doesn't need skip_merge at all, right? The only reason it's being used here is to print that merge is being skipped, You're not actually using the parameter to decide you want to push without merging. Could you please remove it and refactor the code?

scripts/dev/upstream_merge/upstream_merge.py
```diff
@@ -0,0 +1,153 @@
...
+    merge_result = git_obj.merge_branch(
+        f"{git_obj.upstream_repo_name}/{git_obj.upstream_branch}",
+        "Merge latest upstream"
+    )
+
+    if merge_result[0] == 0:
+        diff_output = git_obj.diff()
+        if (git_obj.get_current_commit() == commit_before_merge) or diff_output == (0, ''):
+            return (0, None)
+        return (0, diff_output[1])
```
pratheekshasn: Can you comment here about the difference between `return (0, None)` and `return (0, diff_output[1])`? I'm assuming the first one is for 'OK no changes' and the second one for 'OK', but could you clarify that?

scripts/dev/upstream_merge/upstream_merge.py
pratheekshasn: Basically, if skip_merge is true, don't even call this function instead of calling this function and doing an early return.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,254 @@
...
+        default="scripts/dev/upstream_merge/automation_conf.json"
+    )
+    parser.add_argument("-w", type=str, help="Work item ID", default=None)
+    parser.add_argument("-s", type=str, help="Skip merging with upstream", default=False)
+    args = parser.parse_args()
+    return args
+
+def merge_submodules_with_upstream(
+    conf_file_path, force_checkout, username, upstream_repo_name,
+    merge_branch_name, fork_name, skip_merge
```
pratheekshasn: Same comment as before, instead of using skip_merge here to pass into another function, decide whether to call this function based on the value of the variable in the caller function.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,254 @@
...
+
+def pull_from_base_branch(branch, upstream_url):
+    """
+    Pull the latest changes from the NILRT repository.
+
+    :param branch: The branch name to pull.
+    :param upstream_url: The URL of the upstream repository.
+    :return: A tuple (status_code, message). Returns (0, None) on success.
+    """
+    git_obj = GitRepo()
```
pratheekshasn: Why not add upstream_url, branch_name and repo_path as part of the object itself? And just call git pull?

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,254 @@
...
+    :return: A formatted string containing the PR checklist and work item reference.
+    """
+    return f"""Merge latest from upstream. No conflicts.
+ 
+    #AB{work_item_id}
+    
+    - [x] bitbake packagefeed-ni-core
+    - [x] bitbake packagegroup-ni-desirable
+    - [x] bitbake package-index && bitbake nilrt-base-system-image
+    - [x] Reimaged a cRIO with the new base image and successfully booted it"""
```
pratheekshasn: You'll have to specify that it is a VM.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,254 @@
...
+def get_pr_description(work_item_id):
+    """
+    Generate and return the pull request description text.
+
+    :param work_item_id: The work item ID to reference in the PR description.
+    :return: A formatted string containing the PR checklist and work item reference.
+    """
+    return f"""Merge latest from upstream. No conflicts.
+ 
+    #AB{work_item_id}
```
pratheekshasn: Weren't you going to remove this line if workitem_id was null?

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,254 @@
...
+            merge_report = push_submodules_and_create_PRs(
+                merge_report,
+                json_config_obj.merge_branch_name,
+                "Automated Merge PR",
+                get_pr_description(json_config_obj.work_item_id),
+                json_config_obj.username
+            )
+
+    merge_report["Build and Test"] = build_and_test_details
+
```
pratheekshasn, line 311: The way you update the merge_report is a little inconsistent between the build_and_test stage and the push_and_create_pr stage. You return the result to the caller in build_and_test and append it to the dictionary, whereas in the push_and_create_pr, you update it in the same function. Can you make them both consistent? Choose one or the other.

scripts/dev/utils/push_branch_and_create_pr.py
```diff
@@ -0,0 +1,106 @@
...
+    parser = argparse.ArgumentParser(description="Push and PR script")
+    parser.add_argument("-t", type=str, help="Pull request title", default="Automated Merge PR")
+    parser.add_argument("-u", type=str, help="Username", default=None)
+    parser.add_argument("-br", type=str, help="Branch name", default=None)
+    args = parser.parse_args()
+    return args
+
+if __name__ == "__main__":
+    args = parse_args()
+    git_obj = GitRepo()
```
pratheekshasn: I would set the branch name on the object you have created here, and in push(), just call git_obj.push().

scripts/dev/utils/push_branch_and_create_pr.py
```diff
@@ -0,0 +1,106 @@
...
+
+    status_code, message = push_branch_and_create_pr(git_obj, args.br, args.u, args.t, None)
```
pratheekshasn: Shouldn't this take PR_description as an argument instead of sending a None?

scripts/dev/utils/test.py
```diff
@@ -0,0 +1,187 @@
...
+        {SSH_CONNECTION}:/home/admin"
+        )
+
+def extract_and_install_safemode_image():
+    """
+    Extract and install the safemode image on the target machine using the provided SSH connection.
+    """
+    print("Extracting safemode image on target machine...")
+    return execute_and_stream_cmd_output(f'ssh {SSH_CONNECTION} "tar xf \
+                                         nilrt-safemode-rootfs-x64.tar.gz -C /boot/.safe/"')
```
pratheekshasn: Please create a variable for `nilrt-safemode-rootfs-x64.tar.gz` (and for the runmode tar file) and reuse them in both places. We don't want to debug this script later if the tar file name changes in the future.

pratheekshasn (changes requested):

scripts/dev/upstream_merge/test.py
pratheekshasn: But wouldn't that imply the pipeline should be runnable anywhere? We should pass arguments in the pipeline to things that are configurable, right? This is not really configurable, it's highly specific to this particular machine.

scripts/dev/upstream_merge/README.md
pratheekshasn: Sorry, I meant change "machine" to "machines", as in plural, not remove the word local. I should have been more specific.

scripts/dev/utils/git_repo.py
pratheekshasn: Are you sure appending the current directory is the best way to do this?

And if you do store the absolute path, then you don't need to keep switch to "current directory" every iteration of the loop that runs over each submodule, you can just switch path to git_obj.repo_path

scripts/dev/utils/test.py
CommitSentience: Has been fixed.

scripts/dev/utils/push_branch_and_create_pr.py
CommitSentience: I didn't think user would want to write the whole PR description as a command-line argument.

scripts/dev/utils/push_branch_and_create_pr.py
CommitSentience: Has been updated.

scripts/dev/upstream_merge/upstream_merge_and_test.py
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/upstream_merge.py
CommitSentience: Has been updated.

scripts/dev/upstream_merge/upstream_merge_and_test.py
CommitSentience: I have updated skip_merge such that it is handled in merge_submodules_with_upstream() function in upstream_merge_and_test.py.

scripts/dev/upstream_merge/upstream_merge_and_test.py
CommitSentience: Has been updated.

scripts/dev/upstream_merge/upstream_merge.py
CommitSentience: I have updated skip_merge such that it is handled in merge_submodules_with_upstream() function in upstream_merge_and_test.py.

scripts/dev/upstream_merge/log_and_email_utils.py
CommitSentience: I didn't think anyone would run this script separately.

scripts/dev/upstream_merge/upstream_merge_and_test.py
CommitSentience: Has been updated.

scripts/dev/upstream_merge/upstream_merge_and_test.py
CommitSentience: Has been updated.

scripts/dev/upstream_merge/README.md
CommitSentience: Has been fixed.

scripts/dev/utils/build.py
CommitSentience: I have renamed it.

scripts/dev/utils/push_and_pr.py
CommitSentience: Has been updated.

scripts/dev/utils/test.py
CommitSentience: Has been fixed.

scripts/dev/utils/test.py
CommitSentience: Has been fixed.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,268 @@
...
+def get_pr_description(work_item_id):
+    """
+    Generate and return the pull request description text.
+
+    :param work_item_id: The work item ID to reference in the PR description.
+    :return: A formatted string containing the PR checklist and work item reference.
+    """
+    if work_item_id is not None:
+        return f"""Merge latest from upstream. No conflicts.
+
```
pratheekshasn: [Minor] Can you only append the work_item_id when it's not empty instead of copying the entire description twice? If we decide to change the list of items we need to build in the future, the PR description will be inconsistent based on the value of work_item_id if we don't update it in two places.

pratheekshasn: Also, we're not actually reimaging a cRIO with this workflow, we're only using a VM.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,268 @@
...
+    :param merge_branch_name: The name of the branch to push and create PRs from.
+    :param work_item_id: The work item ID to reference in the PR description.
+    :param username: The GitHub username owning the downstream fork.
+    :return: Updated merge_report dictionary with push and PR results.
+    """
+    push_and_pr_results = {}
+    for git_obj, (status, message) in list(merge_report.items()):
+        if status == 0 and message is not None:
+            current_directory = os.getcwd()
+            os.chdir(git_obj.local_repo)
```
pratheekshasn, line 164: Aren't you storing the absolute path of the local_repo in git_obj.local_repo? Why change back to current directory every iteration then?

scripts/pipelines/clean_build.core-feeds_and_core-images.sh
pratheekshasn: I think this is an option for when we manually build after resolving conflicts, in case of multiple builds.

@Shreejit-03, correct me if I'm wrong.

pratheekshasn (approved):

chaitu236 (commented):
We should we have a feature review first so we're all on same page about the design and goals which would help effectively review this PR.

scripts/dev/upstream_merge/json_config.py
```diff
@@ -0,0 +1,31 @@
...
+
+class JsonConfig:
+    """Handles loading and validating the automation_conf.json configuration file."""
+
+    def __init__(self, automation_conf_path, work_item_id):
+        try:
+            with open(automation_conf_path, "r", encoding="utf-8") as file:
+                config = json.load(file)
+        except (FileNotFoundError, json.JSONDecodeError) as err:
+            print(f"Error loading config file: {err}")
```
chaitu236: The log here prints to stdout rather than stderr. Catching the exception and printing this message and explicitly exiting is not anymore useful than just letting the exception propagate naturally and exit the program.

I think it's just better to get rid of this try/except block entirely.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,259 @@
...
+                upstream_repo_name=upstream_repo_name,
+                fork_name=fork_name,
+                fork_url=fork_url
+            )
+            print(f"\n{git_obj.local_repo.split('/')[-1]}\n")
+            if skip_merge:
+                print(f"\tUpstream Merge has Been Skipped")
+                merge_report[git_obj] = (0, "Upstream Merge has Been Skipped")
+            else:
+                os.chdir(git_obj.local_repo)
```
chaitu236: Spaces before newline here and in other file(s).

image


You could use something like `sed -i 's/[[:space:]]\+$//'` to fix.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,259 @@
+
```
chaitu236: Please run flake8 ( `python3 -m flake8 <file>`) on all the python files and fix all the issues we can. Some of them could be fixed by running black (`python3 -m black <file>`).

scripts/dev/utils/git_commands.py
```diff
@@ -0,0 +1,204 @@
+
+
+"""
+Utility functions for git operations.
```
chaitu236, line 2: Why not use an existing python git library (e.g., GitPython) instead of rolling out our own?

scripts/dev/upstream_merge/upstream_merge.py
```diff
@@ -0,0 +1,146 @@
...
+Workflow for automating the upstream merge process for submodules and repositories.
+"""
+
+import os
+import sys
+# Ensure imports work regardless of current working directory
+PROJECT_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../..'))
+os.chdir(PROJECT_ROOT)
+if PROJECT_ROOT not in sys.path:
+    sys.path.insert(0, PROJECT_ROOT)
```
chaitu236: I don't think this is a good practice and could lead to security issues. e.g., someone could raise a PR with a new file that overrides standard python modules.

You could achieve the same thing by putting the utils dir under `upstream_merge` dir doing `from utils.git_repo import GitRepo`.

Please fix in all files.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,259 @@
...
+    with open(os.path.abspath(conf_file_path), "r", encoding="utf-8") as file:
+        # Read the configuration file - `repos.conf`
+        for line in file:
+            if line.startswith("#"):
+                continue
+            parts = line.split()
+            token = os.environ.get("GH_PAT")
+            if token is None:
+                fork_url = f"https://github.com/{username}/" + parts[0].split("/")[1] + ".git"
```
chaitu236: > parts[0].split("/")[1]

It's better readability wise to create appropriately named local vars that extracts the data you need and use the vars.

e.g.,

```
local_repo = parts[0]
layer_name = local_repo.split("/")[1]

fork_url = f"https://github.com/{username}/{layer_name}.git"
```

Also reduces duplicating `{parts[0].split('/')[-1]` below in the else condition.

scripts/dev/upstream_merge/upstream_merge_and_test.py
```diff
@@ -0,0 +1,259 @@
...
+# Ensure imports work regardless of current working directory
+PROJECT_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../..'))
+os.chdir(PROJECT_ROOT)
+if PROJECT_ROOT not in sys.path:
+    sys.path.insert(0, PROJECT_ROOT)
+
+from scripts.dev.utils.test import install_and_test_image
+from scripts.dev.upstream_merge.json_config import JsonConfig
+from scripts.dev.utils.git_repo import GitRepo
+from scripts.dev.upstream_merge.upstream_merge import merge_upstream
```
chaitu236: `from upstream_merge import merge_upstream` should be sufficient; same for other instances.

chaitu236 (changes requested):
Didn't go through whole PR yet; this is what I have so far.

scripts/dev/utils/git_commands.py
CommitSentience: GitPython doesn’t cover all the cases we need. For example, commands like gh or git send-email aren’t supported, and even for common commands, not all flags are exposed (e.g., some options for git commit --amend or git merge --no-ff/--signoff  are missing). GitPython also raises exceptions when a command fails instead of returning a status code, which makes it harder to uniformly handle errors in scripts. By using git_commands wrapper, we always get a (return_code, output) tuple, giving us full flexibility and consistency across both standard Git commands and less common ones.

chaitu236 (changes requested):
@Shreejit-03 I don't think I'd have any more feedback on these changes. If you can do the following final set of changes that we previously discussed, we can get to merging this PR.

```
The automation_conf.json and the YAML pipeline file will be moved to Azure DevOps.

test.py needs to be updated so that it can be executed on a target device instead of a VM.

Update README.md accordingly to reflect these changes.
```

scripts/dev/upstream_merge/utils/test.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    print("Verifying OS version...")
+    return execute_and_stream_cmd_output(
+        f'ssh -o StrictHostKeyChecking=no {ssh_connection} "cat /etc/os-release"'
+    )
+
+
+if __name__ == "__main__":
+    # Example usage:
+    SSH_CONNECTION = "admin@10.152.9.10"
```
chaitu236: Why are we hardcoding an IP address? We shouldn't have to update this file every time there is a change to target IP. 
This should be handled with an argument passed to `test.py`.

scripts/dev/upstream_merge/automation_conf.json
```diff
@@ -0,0 +1,16 @@
+{
```
chaitu236, line 1: Is this file still needed? I thought you mentioned that this file will be removed.

chaitu236 (changes requested):
I don't see a signed-off tag for the commit message

scripts/dev/upstream_merge/utils/test.py
Shreejit-03: I have removed it.

scripts/dev/upstream_merge/automation_conf.json
Shreejit-03: I have removed the sensitive detail from the configuration file.

chaitu236 (approved):

chaitu236:
Cherry-picked to `nilrt/master/next` - 40ba47db95b037f7200d6885f6c4828d66ae1d29

PR #314, opened 2025-07-25
high-sev-cve-to-json: Update to skip -native packages

usercw88:
# Changes
Update `high-sev-cve-to-json.py` to skip `.*-native` packages
Added catch for non-cve json files

# Testing
* [x] Executed against oe-core/cve folder and verified output

# Process

__Suggested Reviewers:__
* @ni/rtos

### DIFF
--- a/scripts/pipelines/high-sev-cve-to-json.py
+++ b/scripts/pipelines/high-sev-cve-to-json.py
@@ -3,8 +3,8 @@ from argparse import ArgumentParser
 from pathlib import Path
 import json
 import os
+import re
 import sys
-
 
 parser = ArgumentParser(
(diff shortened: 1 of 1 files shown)


scripts/pipelines/high-sev-cve-to-json.py
```diff
@@ -4,7 +4,7 @@
 import json
 import os
 import sys
-
+import re
```
chaitu236, line 6: Can you add `import re` before `import sys` for alphabetical ordering. You might also consider running flake8/black on this file.

chaitu236 (changes requested):

chaitu236 (approved):

PR #325, opened 2025-11-12
Automate Regular x64 stable-rt kernel upgrade

DIVYANSHIKAUSHHIK:
# Changes

This pull request introduces a consolidated script  that fully automates the process of fetching, merging, building, and deploying an updated kernel to a controller. The changes focus on reusing robust utilities from the existing upstream merge framework for improved consistency and maintainability.

https://dev.azure.com/ni/DevCentral/_workitems/edit/3181560
https://dev.azure.com/ni/DevCentral/_workitems/edit/3181567

# Testing

Verified the transfer and installation of the new kernel components on the target controller, followed by a successful reboot. Confirmed the target boots with the new kernel version by executing uname -r on the controller via SSH. Tested the send_email_report function to ensure proper success/failure notifications are sent to the configured recipients

# Process

*TODO: Check boxes that apply to this PR and are complete. Remove boxes that do not apply.*

* [ ] This PR should be cherry-picked to the [`next/`](https://github.com/ni/nilrt/tree/nilrt/master/next) ref.

### DIFF
--- a/scripts/dev/upstream_merge/build_and_install_kernel.py
+++ b/scripts/dev/upstream_merge/build_and_install_kernel.py
@@ -265,7 +265,7 @@ def main():
     global config
     args = parse_args()
     try:
-        config = JsonConfig(automation_conf_path=args.config, work_item_id=None)
+        config = JsonConfig(config_path=args.config, work_item_id=None)
         print(f"[INFO] Loaded configuration from: {args.config}")
         if args.work_dir:
             work_dir = os.path.expandvars(os.path.expanduser(args.work_dir))
--- a/scripts/dev/upstream_merge/build_and_install_kernel.py
+++ b/scripts/dev/upstream_merge/build_and_install_kernel.py
@@ -1,0 +1,291 @@
+#!/usr/bin/env python3
+"""Kernel build, merge (latest stable RT tag), deployment, and notification script.
+Configuration via kernel_build_conf.json.
+"""
+
+import os
+import sys
+import tempfile
+import shutil
+import re
+import argparse
+
+script_dir = os.path.dirname(os.path.abspath(__file__))
+upstream_merge_dir = os.path.join(script_dir, 'dev', 'upstream_merge')
+sys.path.append(upstream_merge_dir)
+from utils.shell_commands import run_command, execute_and_stream_cmd_output
+from utils.git_commands import send_email, git_fetch, git_remote, git_checkout, git_merge, git_reset, git_clean, git_merge_abort, git_tag, git_status, git_diff
+from utils.git_repo import GitRepo
+from utils.toolchain import detect_or_build_cross_compile
+from json_config import JsonConfig
+
+config = None
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Automated kernel build, merge, and deployment script"
+    )
+    parser.add_argument(
+        "-c", "--config",
+        type=str,
+        help="Path to configuration file",
+        default="scripts/kernel_build_conf.json",
+    )
+    parser.add_argument("--skip-merge", action="store_true", help="Skip upstream merge step")
+    parser.add_argument("--skip-packages", action="store_true", help="Skip package installation check")
+    parser.add_argument("--dry-run", action="store_true", help="Show what would be done without executing")
+    parser.add_argument("--work-dir", type=str, help="Override working directory for kernel source", default=None)
+    return parser.parse_args()
+
+# === Utility Functions ===
+def send_email_report(subject, body):
+    tmpfile = tempfile.mktemp()
+    with open(tmpfile, "w") as f:
+        f.write(f"From: {config.email_from}\nTo: {config.email_to}\nSubject: {subject}\n\n{body}\n")
+    try:
+        send_email(to_address=config.email_to, subject=subject, file=tmpfile)
+    finally:
+        os.remove(tmpfile)
+
+# Format merge conflict notification with limited file listing to reduce email size
+def _format_conflict_email(latest_tag, conflicts, git_status_output, target_branch):
+    conflict_list = [c for c in conflicts.splitlines() if c.strip()]
+    max_files = 30
+    shown_files = conflict_list[:max_files]
+    truncated_note = "\n... (truncated)" if len(conflict_list) > max_files else ""
+    # Keep only lines starting with 'U' (unmerged) from status, limit to 20
+    status_lines = [l for l in git_status_output.splitlines() if l.strip().startswith('U')]
+    status_lines = status_lines[:20]
+    status_block = "\n\nGit status (unmerged entries only):\n" + "\n".join(status_lines) if status_lines else ""
+    body = (
+        f"Merge conflict while merging {latest_tag} into {target_branch}\n"
+        f"Total conflicting files: {len(conflict_list)}\n"
+        "Conflicting files:\n" + "\n".join(shown_files) + truncated_note + status_block
+    )
+    return body
+
+# === Step 0: Check Required Packages ===
+def check_required_packages():
+    print("[INFO] Checking required Ubuntu packages...")
+    packages = config.required_packages
+    if not packages:
+        print("[INFO] No required packages specified in configuration.")
+        return
+    missing = []
+    for pkg in packages:
+        status, _ = run_command(f"dpkg -s {pkg}")
+        if status != 0:
+            missing.append(pkg)
+    if missing:
+        print(f"[INFO] Installing missing packages: {', '.join(missing)}")
+        status, output = run_command(f"sudo apt-get update && sudo apt-get install -y {' '.join(missing)}")
+        if status != 0:
+            raise RuntimeError(f"[ERROR] Package install failed: {output}")
+    else:
+        print("[INFO] All required packages are already installed.")
+
+# === Step 1: Detect or Build Toolchain ===
+# Uses shared toolchain detection utility from utils.toolchain module.
+
+# === Step 1.5: Ensure Kernel Source Repository ===
+# Clone kernel repository if missing or not a valid git directory.
+
+def ensure_kernel_source_repository():
+    print("[INFO] Ensuring kernel source repository exists...")
+    if not config.kernel_src_dir:
+        raise RuntimeError("[ERROR] No kernel_src_dir configured in kernel_build section")
+    if not config.repo_url:
+        raise RuntimeError("[ERROR] No repo_url configured for kernel repository")
+    # Convert to absolute path to prevent issues after directory changes
+    config.kernel_src_dir = os.path.abspath(config.kernel_src_dir)
+    kernel_src_dir = config.kernel_src_dir
+    parent_dir = os.path.dirname(kernel_src_dir)
+    if not os.path.exists(parent_dir):
+        print(f"[INFO] Creating parent directory: {parent_dir}")
+        os.makedirs(parent_dir, exist_ok=True)
+    if os.path.exists(kernel_src_dir):
+        if os.path.isdir(os.path.join(kernel_src_dir, '.git')):
+            print(f"[INFO] Kernel source repository present: {kernel_src_dir}")
+            return
+        print(f"[WARNING] {kernel_src_dir} exists but is not a git repository; replacing...")
+        shutil.rmtree(kernel_src_dir)
+    status, output = run_command(f"git clone {config.repo_url} {kernel_src_dir}")
+    if status != 0:
+        raise RuntimeError(f"[ERROR] Kernel repo clone failed: {output}")
+    print(f"[INFO] Cloned kernel repository to: {kernel_src_dir}")
+
+# === Step 2: Robust Upstream Merge with Conflict Handling ===
+def run_upstream_merge_script():
+    print("[INFO] Running upstream merge script...")
+    ensure_kernel_source_repository()
+    if not os.path.exists(config.kernel_src_dir):
+        raise RuntimeError(f"[ERROR] Kernel source directory {config.kernel_src_dir} missing after clone")
+    original_cwd = os.getcwd()
+    kernel_parent_dir = os.path.dirname(config.kernel_src_dir)
+    os.chdir(kernel_parent_dir)
+    try:
+        kernel_version = config.target_branch.split('/')[-1] if config.target_branch else "6.12"
+        git_obj = GitRepo(
+            local_repo=os.path.basename(config.kernel_src_dir),
+            upstream_repo_url=config.stable_rt_remote,
+            upstream_branch=f"linux-{kernel_version}.y-rt",
+            local_base_branch=config.target_branch,
+            upstream_repo_name="stable-rt",
+            fork_name="origin",
+            fork_url=config.repo_url
+        )
+        os.chdir(config.kernel_src_dir)
+        try:
+            git_merge_abort()
+        except:
+            pass
+        git_reset(hard=True)
+        git_clean(force=True, directories=True, ignored_files=True)
+        status, _ = git_fetch()
+        if status != 0:
+            raise RuntimeError("Failed to fetch remotes")
+        status, _ = run_command(f"git checkout -B {config.target_branch} origin/{config.target_branch}")
+        if status != 0:
+            raise RuntimeError("Failed to checkout target branch")
+        status, remotes = git_remote()
+        if status == 0 and "stable-rt" not in remotes:
+            git_obj.add_remote("stable-rt", config.stable_rt_remote)
+        git_fetch("stable-rt", "--tags")
+        status, tags = git_tag(list_pattern=f"v{kernel_version}.*-rt*")
+        if status != 0 or not tags:
+            print(f"[WARNING] No v{kernel_version}-rt tags; skipping merge")
+            return
+        clean_tags = [t for t in tags.splitlines() if re.match(rf'^v{re.escape(kernel_version)}\.\d+(?:\.\d+)?-rt\d+$', t)]
+        if not clean_tags:
+            print(f"[WARNING] No clean v{kernel_version}-rt release tags; skipping merge")
+            return
+        latest_tag = sorted(clean_tags, key=lambda t: list(map(int, re.findall(r'\d+', t))))[-1]
+        print(f"[INFO] Latest RT tag: {latest_tag}")
+        merge_result = git_obj.merge_branch(latest_tag, f"Merge latest upstream {latest_tag}")
+        if merge_result[0] != 0:
+            status, conflicts = git_diff(name_only=True, diff_filter="U")
+            if status == 0 and conflicts:
+                status, git_status_output = git_status()
+                concise_body = _format_conflict_email(latest_tag, conflicts, git_status_output, config.target_branch)
+                send_email_report(
+                    f"Merge conflict report: {config.target_branch}",
+                    concise_body
+                )
+                raise RuntimeError("[ERROR] Merge conflicts detected; email sent")
+        else:
+            print("[INFO] Merge successful")
+    finally:
+        os.chdir(original_cwd)
+
+# === Step 3: Build and Deploy Kernel ===
+def build_and_deploy_kernel(args):
+    try:
+        if not args.skip_packages:
+            check_required_packages()
+        CROSS_COMPILE = detect_or_build_cross_compile(config, script_dir)
+        env = os.environ.copy()
+        if config.arch:
+            env["ARCH"] = config.arch
+        env["CROSS_COMPILE"] = CROSS_COMPILE
+        if not args.skip_merge:
+            run_upstream_merge_script()
+        else:
+            ensure_kernel_source_repository()
+        kernel_src_dir = config.kernel_src_dir
+        print("[INFO] Cleaning previous builds...")
+        if not args.dry_run:
+            status, output = run_command("make mrproper", cwd=kernel_src_dir, env=env)
+            if status != 0:
+                raise RuntimeError(output)
+        kernel_config_target = config.kernel_config or "defconfig"
+        print(f"[INFO] Creating kernel configuration: {kernel_config_target}...")
+        if not args.dry_run:
+            status, output = run_command(f"make {kernel_config_target}", cwd=kernel_src_dir, env=env)
+            if status != 0:
+                raise RuntimeError(output)
+        make_jobs = config.make_jobs or "$(nproc)"
+        if make_jobs == "$(nproc)":
+            make_jobs = str(os.cpu_count())
+        print(f"[INFO] Building kernel and modules with {make_jobs} jobs...")
+        if not args.dry_run:
+            status, output = run_command(f"make -j{make_jobs} bzImage modules", cwd=kernel_src_dir, env=env)
+            if status != 0:
+                raise RuntimeError(output)
+        temp_modules_dir = config.temp_modules_dir or os.path.join(kernel_src_dir, "tmp-modules")
+        print(f"[INFO] Staging modules in: {temp_modules_dir}...")
+        if not args.dry_run:
+            if os.path.exists(temp_modules_dir):
+                shutil.rmtree(temp_modules_dir)
+            os.makedirs(temp_modules_dir, exist_ok=True)
+            status, output = run_command(f"make modules_install INSTALL_MOD_PATH={temp_modules_dir}", cwd=kernel_src_dir, env=env)
+            if status != 0:
+                raise RuntimeError(output)
+        kernel_version = run_command("make -s kernelrelease", cwd=kernel_src_dir, env=env)[1] if not args.dry_run else "DRY-RUN-VERSION"
+        target_host = config.kernel_target_host or config.rt_target_IP
+        target_user = config.kernel_target_user or "admin"
+        if not target_host:
+            raise RuntimeError("[ERROR] No target host configured")
+        print(f"[INFO] Copying kernel to target {target_host}...")
+        if not args.dry_run:
+            status, output = run_command(f"scp {kernel_src_dir}/arch/x86/boot/bzImage {target_user}@{target_host}:/boot/bzImage-{kernel_version}")
+            if status != 0:
+                raise RuntimeError(output)
+        print("[INFO] Backing up existing kernel on target (if needed)...")
+        if not args.dry_run:
+            status, output = run_command(f"ssh {target_user}@{target_host} 'if [ -f /boot/runmode/bzImage ] && [ ! -h /boot/runmode/bzImage ]; then mv /boot/runmode/bzImage /boot/runmode/bzImage-$(uname -r); else echo Skipping backup; fi'")
+            if status != 0:
+                print(f"[WARNING] Backup step reported: {output}")
+        print("[INFO] Updating bootloader symlink...")
+        if not args.dry_run:
+            status, output = run_command(f"ssh {target_user}@{target_host} 'ln -sf bzImage-{kernel_version} /boot/runmode/bzImage'")
+            if status != 0:
+                raise RuntimeError(output)
+        print("[INFO] Copying modules to target...")
+        if not args.dry_run:
+            status, output = execute_and_stream_cmd_output(f"tar cz -C {temp_modules_dir} lib | ssh {target_user}@{target_host} tar xz -C /")
+            if status != 0:
+                raise RuntimeError(f"[ERROR] Module transfer failed: {output}")
+        print("[INFO] Rebooting target...")
+        if not args.dry_run:
+            status, output = execute_and_stream_cmd_output(f"ssh {target_user}@{target_host} 'reboot' || true")
+            if status != 0:
+                print(f"[WARNING] Reboot command returned non-zero: {output}")
+        if not args.dry_run:
+            send_email_report("[SUCCESS] Kernel Build+Install", f"Kernel build and deploy succeeded on {target_host}, running version {kernel_version}")
+        else:
+            print("[INFO] Dry run completed successfully")
+    except Exception as e:
+        print(f"[ERROR] {e}")
+        if not args.dry_run:
+            send_email_report("[FAILURE] Kernel Build+Install", f"Kernel build/deploy failed: {e}")
+        sys.exit(1)
+
+# === Main ===
+def main():
+    global config
+    args = parse_args()
+    try:
+        config = JsonConfig(automation_conf_path=args.config, work_item_id=None)
+        print(f"[INFO] Loaded configuration from: {args.config}")
+        if args.work_dir:
+            work_dir = os.path.expandvars(os.path.expanduser(args.work_dir))
+            config.kernel_src_dir = os.path.join(work_dir, "linux")
+            print(f"[INFO] Using work directory override: {args.work_dir}")
+            print(f"[INFO] Kernel source path: {config.kernel_src_dir}")
+        if getattr(config, 'kernel_src_dir', None):
+            # Convert to absolute path to prevent issues after directory changes
+            config.kernel_src_dir = os.path.abspath(config.kernel_src_dir)
+            target_host = config.kernel_target_host or config.rt_target_IP or "not configured"
+            arch = config.arch or "not configured"
+            branch = config.target_branch or "not configured"
+            print(f"[INFO] Target: {target_host}, Arch: {arch}, Branch: {branch}")
+            print(f"[INFO] Absolute kernel source path: {config.kernel_src_dir}")
+        else:
+            print("[WARNING] kernel_build section missing or incomplete")
+    except Exception as e:
+        print(f"[ERROR] Failed to load configuration: {e}")
+        sys.exit(1)
+    build_and_deploy_kernel(args)
+
+if __name__ == "__main__":
+    main()
--- a/scripts/dev/upstream_merge/build_and_install_kernel.py
+++ b/scripts/dev/upstream_merge/build_and_install_kernel.py
@@ -90,39 +90,52 @@ def check_required_packages():
 # === Step 1.5: Ensure Kernel Source Repository ===
 # Clone kernel repository if missing or not a valid git directory.
 
-def ensure_kernel_source_repository():
+def ensure_kernel_source_repository(args):
     print("[INFO] Ensuring kernel source repository exists...")
+    
+    # Use default kernel source directory if not configured
     if not config.kernel_src_dir:
         raise RuntimeError("[ERROR] No kernel_src_dir configured in kernel_build section")
-    if not config.repo_url:
-        raise RuntimeError("[ERROR] No repo_url configured for kernel repository")
+    
+    # Use NI Linux repository as default - this is the primary repo for NILRT
+    repo_url = "https://github.com/ni/linux.git"
     # Convert to absolute path to prevent issues after directory changes
     config.kernel_src_dir = os.path.abspath(config.kernel_src_dir)
     kernel_src_dir = config.kernel_src_dir
     parent_dir = os.path.dirname(kernel_src_dir)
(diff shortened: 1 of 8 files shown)


scripts/dev/upstream_merge/utils/toolchain.py
```diff
@@ -0,0 +1,83 @@
+"""Toolchain detection/build helper: auto-detect existing repo first (ancestor walk),
+ignore nilrt_root hint if already inside a clone; only clone when nothing usable exists."""
+import os
+from .shell_commands import run_command
+from .git_commands import git_clone
+
+BUILD_SCRIPT_REL = os.path.join('scripts', 'pipelines', 'build.toolchain.sh')
```
pratheekshasn, line 7: Any reason `BUILD_SCRIPT_REL` is made global _and_ passed as a variable?

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+    )
+    parser.add_argument(
+        "-c", "--config",
+        type=str,
+        help="Path to configuration file",
+        default="scripts/kernel_build_conf.json",
+    )
+    parser.add_argument("--skip-merge", action="store_true", help="Skip upstream merge step")
+    parser.add_argument("--skip-packages", action="store_true", help="Skip package installation check")
+    parser.add_argument("--dry-run", action="store_true", help="Show what would be done without executing")
```
pratheekshasn, line 36: Where and how is `--dry-run` used? I don't see any other place in this whole script where it is being used.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+    parser.add_argument("--work-dir", type=str, help="Override working directory for kernel source", default=None)
```
pratheekshasn, line 37: Same as before. Where and how is `--work-dir` being used?

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+    parser = argparse.ArgumentParser(
+        description="Automated kernel build, merge, and deployment script"
+    )
+    parser.add_argument(
+        "-c", "--config",
+        type=str,
+        help="Path to configuration file",
+        default="scripts/kernel_build_conf.json",
+    )
+    parser.add_argument("--skip-merge", action="store_true", help="Skip upstream merge step")
```
pratheekshasn, line 34: Is `--skip-merge` being used anywhere at all?

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+    parser.add_argument("--skip-packages", action="store_true", help="Skip package installation check")
```
pratheekshasn, line 35: Is `--skip-packages` being used anywhere at all?

scripts/dev/upstream_merge/utils/toolchain.py
```diff
@@ -0,0 +1,83 @@
...
+        else:
+            print(f'[WARNING] No SDK installer found in {sdk_dir}')
+    else:
+        print(f'[WARNING] SDK directory {sdk_dir} does not exist')
+
+    gcc_binary = config.get_toolchain_gcc_path()
+    if gcc_binary and os.path.isfile(gcc_binary):
+        print(f"[INFO] Successfully built CROSS_COMPILE: {config.toolchain_prefix}")
+        return config.toolchain_prefix
+    raise RuntimeError('[ERROR] Toolchain build completed, but compiler still missing.')
```
pratheekshasn: You might want to change this error message to say that the compiler is missing in the provided path, instead of just saying it's missing.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+        print("[INFO] No required packages specified in configuration.")
+        return
+    missing = []
+    for pkg in packages:
+        status, _ = run_command(f"dpkg -s {pkg}")
+        if status != 0:
+            missing.append(pkg)
+    if missing:
+        print(f"[INFO] Installing missing packages: {', '.join(missing)}")
+        status, output = run_command(f"sudo apt-get update && sudo apt-get install -y {' '.join(missing)}")
```
pratheekshasn, line 81: Just curious, what was the error you were getting before when you said you had to use `subprocess.run()` instead of the existing `run_command()` when you had to use `||` or `&&` in the command? I seem to remember you mentioning `shlex.split()` as a reason for that. How did you fix that error?

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+# === Step 1: Detect or Build Toolchain ===
+# Uses shared toolchain detection utility from utils.toolchain module.
+
+# === Step 1.5: Ensure Kernel Source Repository ===
+# Clone kernel repository if missing or not a valid git directory.
+
+def ensure_kernel_source_repository():
+    print("[INFO] Ensuring kernel source repository exists...")
+    if not config.kernel_src_dir:
+        raise RuntimeError("[ERROR] No kernel_src_dir configured in kernel_build section")
```
pratheekshasn, line 98: You should choose a default path for cloning the kernel source repo if it's not provided in the config, that would be better instead of failing.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+    if not config.repo_url:
+        raise RuntimeError("[ERROR] No repo_url configured for kernel repository")
```
pratheekshasn: We already know the `repo_url`, right? So you should use that value when it's not provided in the config. Also, why does this need to be in the config at all? It's unlikely that the URL is going to change frequently.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+    # Convert to absolute path to prevent issues after directory changes
+    config.kernel_src_dir = os.path.abspath(config.kernel_src_dir)
```
pratheekshasn, line 103: How exactly does it create an absolute path for the relative path that's given in the config? As in, once it creates the absolute path, what value does it have? `$HOME/<config_relative_path>` or `<current_dir>/config_relative_path`?

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+    kernel_src_dir = config.kernel_src_dir
+    parent_dir = os.path.dirname(kernel_src_dir)
+    if not os.path.exists(parent_dir):
+        print(f"[INFO] Creating parent directory: {parent_dir}")
```
pratheekshasn: Follow up question to the absolute/relative path question about the `kernel_src_dir`: Why would the parent directory not exist? If it doesn't exist, how is `os.path.dirname()` going to fetch that parent path?

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+
+# === Step 2: Robust Upstream Merge with Conflict Handling ===
+def run_upstream_merge_script():
+    print("[INFO] Running upstream merge script...")
+    ensure_kernel_source_repository()
+    if not os.path.exists(config.kernel_src_dir):
+        raise RuntimeError(f"[ERROR] Kernel source directory {config.kernel_src_dir} missing after clone")
+    original_cwd = os.getcwd()
+    kernel_parent_dir = os.path.dirname(config.kernel_src_dir)
+    os.chdir(kernel_parent_dir)
```
pratheekshasn: Why change to `kernel_parent_dir`, do nothing with it, and then change to kernel_src_dir` in line 137?

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+        try:
+            git_merge_abort()
+        except:
+            pass
+        git_reset(hard=True)
+        git_clean(force=True, directories=True, ignored_files=True)
+        status, _ = git_fetch()
+        if status != 0:
+            raise RuntimeError("Failed to fetch remotes")
+        status, _ = run_command(f"git checkout -B {config.target_branch} origin/{config.target_branch}")
```
pratheekshasn: You might want to use the existing `git_checkout` function present in `git_commands.py`, that way all checkouts will be consistent. In case we need to make a change to it later, it will be changed in one single location and you need not hunt through the entire codebase looking for where `git checkout ...`  is being used.

scripts/dev/upstream_merge/build_and_install_kernel.py
pratheekshasn: I see you have a typo in your code. The argument you're parsing here is `skip-packages`, but the name you're using in the code further below is `skip_packages`. Please test the code when you make changes and push it.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+"""
+
+import os
+import sys
+import tempfile
+import shutil
+import re
+import argparse
+
+script_dir = os.path.dirname(os.path.abspath(__file__))
```
pratheekshasn, line 13: `script_dir` variable wouldn't change, at least not within the lifetime of the program. The typical naming convention would be to use all upper case.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+            print("[INFO] Merge successful")
+    finally:
+        os.chdir(original_cwd)
+
+# === Step 3: Build and Deploy Kernel ===
+def build_and_deploy_kernel(args):
+    try:
+        if not args.skip_packages:
+            check_required_packages()
+        CROSS_COMPILE = detect_or_build_cross_compile(config, script_dir)
```
pratheekshasn: Aren't `config` and `script_dir` already global? Why pass them to another function as parameters?

pratheekshasn, line 206: You should probably rename `CROSS_COMPILE` to lower case, typically upper case naming is reserved for constants.

scripts/dev/upstream_merge/build_and_install_kernel.py
pratheekshasn: Same comment as `skip-packages`, you have a typo in the name of arguments you're parsing, and then later using it in the code.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+            return
+        print(f"[WARNING] {kernel_src_dir} exists but is not a git repository; replacing...")
+        shutil.rmtree(kernel_src_dir)
+    status, output = run_command(f"git clone {config.repo_url} {kernel_src_dir}")
+    if status != 0:
+        raise RuntimeError(f"[ERROR] Kernel repo clone failed: {output}")
+    print(f"[INFO] Cloned kernel repository to: {kernel_src_dir}")
+
+# === Step 2: Robust Upstream Merge with Conflict Handling ===
+def run_upstream_merge_script():
```
pratheekshasn: Rather than call this `run_upstream_merge_script()`, you should name it something along the lines of `merge_upstream_kernel` or something similar.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+            check_required_packages()
+        CROSS_COMPILE = detect_or_build_cross_compile(config, script_dir)
+        env = os.environ.copy()
+        if config.arch:
+            env["ARCH"] = config.arch
+        env["CROSS_COMPILE"] = CROSS_COMPILE
+        if not args.skip_merge:
+            run_upstream_merge_script()
+        else:
+            ensure_kernel_source_repository()
```
pratheekshasn: Why do we need to ensure that the kernel source directory exists if we do decide to skip_merge?

So the reason we use the skip_merge option is when there are conflicts, and we pause the build machine and intervene. It would have failed because of the conflicts, so we resolve the conflicts manually, and then use the skip_merge option of the script so that it doesn't fetch again and cause conflicts again. Ideally, when you run with the skip_merge option, it will be the second time you would be running the script. So the kernel source directory is going to be present.

But if you have another reason why this check is required, please reply below, I might have missed some work flow.

scripts/dev/upstream_merge/build_and_install_kernel.py
pratheekshasn: Typo: `dry-run` vs `dry_run`. I know it's hard to generate conflicts and test out the skip-merge functionality, but please do test the dry-run functionality whenever you make any changes to the code.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+        kernel_src_dir = config.kernel_src_dir
+        print("[INFO] Cleaning previous builds...")
+        if not args.dry_run:
+            status, output = run_command("make mrproper", cwd=kernel_src_dir, env=env)
```
pratheekshasn, line 218: It's better to declare these commands as constants on the top of the module, and use those constants instead of string literals.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+            if status != 0:
+                raise RuntimeError(output)
+        kernel_config_target = config.kernel_config or "defconfig"
```
pratheekshasn, line 221: Why run `defconfig` if the config file doesn't provide that? Wouldn't that mean it wouldn't have any of the settings explicitly made by NI?

pratheekshasn, line 221: Isn't `config.kernel_config` just the command you need to run to configure the kernel that you're going to build? The variable name doesn't really convey that.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+            make_jobs = str(os.cpu_count())
+        print(f"[INFO] Building kernel and modules with {make_jobs} jobs...")
+        if not args.dry_run:
+            status, output = run_command(f"make -j{make_jobs} bzImage modules", cwd=kernel_src_dir, env=env)
+            if status != 0:
+                raise RuntimeError(output)
+        temp_modules_dir = config.temp_modules_dir or os.path.join(kernel_src_dir, "tmp-modules")
+        print(f"[INFO] Staging modules in: {temp_modules_dir}...")
+        if not args.dry_run:
+            if os.path.exists(temp_modules_dir):
```
pratheekshasn, line 238: Would the temp_modules_dir exist after running `make mrproper`?

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+        if not target_host:
+            raise RuntimeError("[ERROR] No target host configured")
+        print(f"[INFO] Copying kernel to target {target_host}...")
+        if not args.dry_run:
+            status, output = run_command(f"scp {kernel_src_dir}/arch/x86/boot/bzImage {target_user}@{target_host}:/boot/bzImage-{kernel_version}")
+            if status != 0:
+                raise RuntimeError(output)
+        print("[INFO] Backing up existing kernel on target (if needed)...")
+        if not args.dry_run:
+            status, output = run_command(f"ssh {target_user}@{target_host} 'if [ -f /boot/runmode/bzImage ] && [ ! -h /boot/runmode/bzImage ]; then mv /boot/runmode/bzImage /boot/runmode/bzImage-$(uname -r); else echo Skipping backup; fi'")
```
pratheekshasn, line 256: You might want to run the backup step before actually copying the kernel over to the target.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,291 @@
...
+            status, output = execute_and_stream_cmd_output(f"tar cz -C {temp_modules_dir} lib | ssh {target_user}@{target_host} tar xz -C /")
+            if status != 0:
+                raise RuntimeError(f"[ERROR] Module transfer failed: {output}")
+        print("[INFO] Rebooting target...")
+        if not args.dry_run:
+            status, output = execute_and_stream_cmd_output(f"ssh {target_user}@{target_host} 'reboot' || true")
+            if status != 0:
+                print(f"[WARNING] Reboot command returned non-zero: {output}")
+        if not args.dry_run:
+            send_email_report("[SUCCESS] Kernel Build+Install", f"Kernel build and deploy succeeded on {target_host}, running version {kernel_version}")
```
pratheekshasn, line 275: You're sending a report here right after reboot, and reporting a success before even testing it. You need to let it reboot, wait till it comes back up, run `uname -r` and verify that it matches the latest tag you have fetched.

scripts/dev/upstream_merge/json_config.py
```diff
@@ -9,8 +9,8 @@ class JsonConfig:
     """Handles loading and validating the automation_conf.json
         configuration file."""
 
-    def __init__(self, automation_conf_path, work_item_id):
-        with open(automation_conf_path, "r", encoding="utf-8") as file:
+    def __init__(self, config_path, work_item_id):
+        with open(config_path, "r", encoding="utf-8") as file:
             config = json.load(file)
         self.work_item_id = work_item_id
         self.nilrt_branch = config.get("nilrt_branch")
```
pratheekshasn, line 16: You don't have this in your config file. What happens then? Does it just have an empty string without throwing an error?

scripts/dev/upstream_merge/json_config.py
```diff
@@ -27,3 +27,45 @@ def __init__(self, automation_conf_path, work_item_id):
...
+        self.temp_modules_dir = self._expand_path(kernel_config.get("temp_modules_dir"))
+        self.toolchain_prefix = self._expand_path(kernel_config.get("toolchain_prefix"))
+        self.merge_workdir = self._expand_path(kernel_config.get("merge_workdir"))
+        self.repo_url = kernel_config.get("repo_url")
+        self.target_branch = kernel_config.get("target_branch")
+        self.stable_rt_remote = kernel_config.get("stable_rt_remote")
+        self.nilrt_root = self._expand_path(kernel_config.get("nilrt_root"))
+        self.required_packages = kernel_config.get("required_packages", [])
+        self.make_jobs = kernel_config.get("make_jobs", "$(nproc)")
+        self.kernel_config = kernel_config.get("kernel_config", "defconfig")
```
pratheekshasn, line 46: Probably shouldn't use `defconfig` as the default value, it wouldn't have NI specific kernel configurations. And I _think_ it will fail to boot the target when you use `defconfig`, if I remember correctly.

scripts/dev/upstream_merge/kernel_build_conf.json
```diff
@@ -0,0 +1,22 @@
+{
```
pratheekshasn, line 1: This is fine for now, but you'll need to move these variables to the pipeline later.

scripts/dev/upstream_merge/kernel_build_conf.json
```diff
@@ -0,0 +1,22 @@
...
+    "toolchain_prefix": "/usr/local/oecore-x86_64/sysroots/x86_64-nilrtsdk-linux/usr/bin/x86_64-nilrt-linux/x86_64-nilrt-linux-",
+    "repo_url": "https://github.com/ni/linux.git",
+    "target_branch": "nilrt/master/6.12",
+    "stable_rt_remote": "https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-stable-rt.git",
+    "nilrt_root": "nilrt-sdk",
+    "required_packages": [
+      "bc", "bison", "flex", "gcc", "git",
+      "kmod", "libelf-dev", "libssl-dev", "make", "u-boot-tools"
+    ],
+    "make_jobs": "$(nproc)",
```
pratheekshasn, line 19: Is this something that could change frequently? Otherwise it doesn't make sense to add it in the config.

scripts/dev/upstream_merge/kernel_build_conf.json
```diff
@@ -0,0 +1,22 @@
...
+    "target_host": "10.152.8.54",
+    "target_user": "admin",
+    "arch": "x86_64",
+    "temp_modules_dir": "nilrt-kernel-build/linux/tmp-glibc/modules",
+    "toolchain_prefix": "/usr/local/oecore-x86_64/sysroots/x86_64-nilrtsdk-linux/usr/bin/x86_64-nilrt-linux/x86_64-nilrt-linux-",
+    "repo_url": "https://github.com/ni/linux.git",
+    "target_branch": "nilrt/master/6.12",
+    "stable_rt_remote": "https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-stable-rt.git",
+    "nilrt_root": "nilrt-sdk",
+    "required_packages": [
```
pratheekshasn, line 15: Are these values likely to change frequently? Otherwise it doesn't make sense to add to a config file.

scripts/dev/upstream_merge/kernel_build_conf.json
```diff
@@ -0,0 +1,22 @@
+{
+  "email_from": "dkaushik@emerson.com",
+  "email_to": "dkaushik@emerson.com",
+  "kernel_build": {
+    "kernel_src_dir": "nilrt-kernel-build/linux",
+    "target_host": "10.152.8.54",
+    "target_user": "admin",
+    "arch": "x86_64",
+    "temp_modules_dir": "nilrt-kernel-build/linux/tmp-glibc/modules",
```
pratheekshasn, line 9: Shouldn't this just be `./tmp-modules` according to the [README](https://github.com/ni/nilrt/blob/nilrt/master/scarthgap/docs/README.kernel.md)? You should be able to build the path with the kernel source directory in your code, instead of in your config.

scripts/dev/upstream_merge/utils/git_commands.py
```diff
@@ -262,6 +275,104 @@ def git_pull_request(
...
+    if force:
+        command += " -f"
+    if directories:
+        command += " -d"
+    if ignored_files:
+        command += " -x"
+    return run_command(command, capture_output)
+
+
+def git_merge_abort(
```
pratheekshasn, line 323: Why not modify the existing git_merge function instead of adding a new one? That way we have lesser number of merge related functions to keep track of.

scripts/dev/upstream_merge/utils/toolchain.py
```diff
@@ -0,0 +1,83 @@
...
+      1. Ancestor of script_dir containing build.toolchain.sh (preferred)
+      2. Provided hint (config.nilrt_root) if ancestor not found
+      3. Clone into expanded hint (or ~/nilrt-sdk if empty) if still missing
+    """
+    gcc_binary = config.get_toolchain_gcc_path()
+    if gcc_binary and os.path.isfile(gcc_binary):
+        print(f"[INFO] Found CROSS_COMPILE: {config.toolchain_prefix}")
+        return config.toolchain_prefix
+    if not config.toolchain_prefix:
+        raise RuntimeError('[ERROR] No toolchain_prefix configured')
```
pratheekshasn, line 34: Wouldn't this kill the script if the value is not provided? Since we have a proper default path for where the toolchain should be installed, you should use that path instead of terminating early.

scripts/dev/upstream_merge/utils/toolchain.py
```diff
@@ -0,0 +1,83 @@
...
+    if status != 0:
+        raise RuntimeError('[ERROR] Toolchain build script failed')
+
+    sdk_dir = config.get_sdk_dir()
+    if sdk_dir and os.path.exists(sdk_dir):
+        installers = [f for f in os.listdir(sdk_dir) if f.endswith('.sh')]
+        if installers:
+            installer = os.path.join(sdk_dir, installers[0])
+            print('[INFO] Installing SDK to /usr/local/oecore-x86_64 ...')
+            status, _ = run_command(f"sudo bash {installer} -d /usr/local/oecore-x86_64 -y")
```
pratheekshasn, line 71: You have hardcoded the path `/usr/local/oecore-x86_64` here. What if the config file has provided a custom value?

scripts/kernel_build_conf.json
```diff
@@ -0,0 +1,19 @@
+{
```
pratheekshasn, line 1: This if fine for now, but you may need to move some of these values to the pipeline in the future.

scripts/kernel_build_conf.json
```diff
@@ -0,0 +1,19 @@
...
+  "email_from": "dkaushik@emerson.com",
+  "email_to": "dkaushik@emerson.com",
+  "kernel_build": {
+    "kernel_src_dir": "nilrt-kernel-build/linux",
+    "target_host": "10.152.8.54",
+    "target_user": "admin",
+    "arch": "x86_64",
+    "temp_modules_dir": "nilrt-kernel-build/linux/tmp-glibc/modules",
```
pratheekshasn, line 9: Isn't this just `.tmp-modules` according to the README? Typically you would build this path with the kernel source directory in your code, and keep the config minimal.

scripts/kernel_build_conf.json
```diff
@@ -0,0 +1,19 @@
...
+    "repo_url": "https://github.com/ni/linux.git",
+    "target_branch": "nilrt/master/6.12",
+    "stable_rt_remote": "https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-stable-rt.git",
+    "nilrt_root": "nilrt-sdk",
+    "toolchain_prefix": "/usr/local/oecore-x86_64/sysroots/x86_64-nilrtsdk-linux/usr/bin/x86_64-nilrt-linux/x86_64-nilrt-linux-",
+    "required_packages": ["bc", "bison", "flex", "gcc", "git", "kmod", "libelf-dev", "libssl-dev", "make", "u-boot-tools"],
+    "make_jobs": "$(nproc)",
```
pratheekshasn, line 15: This is unlikely to change frequently, right? If so, you can just add the command in the code instead of the config.

scripts/kernel_build_conf.json
```diff
@@ -0,0 +1,19 @@
...
+    "target_host": "10.152.8.54",
+    "target_user": "admin",
+    "arch": "x86_64",
+    "temp_modules_dir": "nilrt-kernel-build/linux/tmp-glibc/modules",
+    "repo_url": "https://github.com/ni/linux.git",
+    "target_branch": "nilrt/master/6.12",
+    "stable_rt_remote": "https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-stable-rt.git",
+    "nilrt_root": "nilrt-sdk",
+    "toolchain_prefix": "/usr/local/oecore-x86_64/sysroots/x86_64-nilrtsdk-linux/usr/bin/x86_64-nilrt-linux/x86_64-nilrt-linux-",
+    "required_packages": ["bc", "bison", "flex", "gcc", "git", "kmod", "libelf-dev", "libssl-dev", "make", "u-boot-tools"],
```
pratheekshasn, line 14: This is unlikely to change frequently, right? If so, you can just add the list of packages in the code instead of the config.

pratheekshasn (changes requested):
Please do test options like skip-merge and dry-run before you push the code. It may be difficult to generate conflicts to test the skip-merge option, but running the dry-run option at least should be sufficient.

scripts/dev/upstream_merge/build_and_install_kernel.py
DIVYANSHIKAUSHHIK: Command line argument: `--dry-run` (with hyphen)
Python variable: `args.dry_run` (with underscore)
 
This is correct - argparse automatically converts hyphens to underscores in the variable names. The naming convention (`--dry-run` → `args.dry_run`) is correct as per argparse standards.

scripts/dev/upstream_merge/build_and_install_kernel.py
DIVYANSHIKAUSHHIK: The `--work-dir` functionality  - provides a clean way to override the default kernel source location for testing or different build environments.. However, like with `--dry-run`, the naming convention follows argparse standards:
 Command line: `--work-dir` (hyphen)
Python variable: `args.work_dir` (underscore)
 
This is the standard behavior for argparse - it automatically converts hyphens to underscores in variable names.

scripts/dev/upstream_merge/build_and_install_kernel.py
DIVYANSHIKAUSHHIK: --skip-packages: When used, skips the package installation check
--skip-merge: When used, skips the upstream merge but still ensures the kernel source repository exists, The argparse behavior of converting hyphens to underscores in attribute names is the standard Python convention

scripts/dev/upstream_merge/build_and_install_kernel.py
DIVYANSHIKAUSHHIK: --skip-packages: When used, skips the package installation check
--skip-merge: When used, skips the upstream merge but still ensures the kernel source repository exists, The argparse behavior of converting hyphens to underscores in attribute names is the standard Python convention

nilrt-kernel-build/linux
pratheekshasn, line 1: You should not be adding this to the `nilrt` repo. You can clone the repo within this path (if and only if necessary) for merging the new kernel code, but you should not be submitting this change to your PR.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,312 @@
...
+            raise RuntimeError(f"[ERROR] Kernel repo clone failed: {output}")
+        print(f"[INFO] Cloned kernel repository to: {kernel_src_dir}")
+    else:
+        print(f"[INFO] Would clone kernel repository from {repo_url} to: {kernel_src_dir}")
+
+# === Step 2: Robust Upstream Merge with Conflict Handling ===
+def run_upstream_merge_script(args):
+    print("[INFO] Running upstream merge script...")
+    ensure_kernel_source_repository(args)
+    if not args.dry_run and not os.path.exists(config.kernel_src_dir):
```
pratheekshasn, line 133: `if not args.dry_run and not os.path.exists(config.kernel_src_dir):`
`        raise RuntimeError(f"[ERROR] Kernel source directory {config.kernel_src_dir} missing after clone")`

Is this even necessary after having called `ensure_kernel_source_repository()`?

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,312 @@
...
+        raise RuntimeError(f"[ERROR] Kernel source directory {config.kernel_src_dir} missing after clone")
+    original_cwd = os.getcwd()
+    if not args.dry_run:
+        kernel_parent_dir = os.path.dirname(config.kernel_src_dir)
+        os.chdir(kernel_parent_dir)
+    try:
+        kernel_version = config.target_branch.split('/')[-1] if config.target_branch else "6.12"
```
pratheekshasn, line 140: It might be safer to not have to fallback to 6.12 when the config file doesn't specify it. If someone forgets to update the config file when they run this later, we shouldn't keep using 6.12 even though there may be later versions present.

scripts/dev/upstream_merge/utils/shell_commands.py
```diff
@@ -8,22 +8,24 @@
...
     :param command: Shell command as a string.
     :param capture_output: Whether to capture the output.
+    :param cwd: Working directory for command execution.
+    :param env: Environment variables for command execution.
     :return: (return_code, output) - return code and
                                     output string (or None if not captured).
     """
     print(command)
     logger.info("Running command: %s", command)
-    formatted_command = shlex.split(command)
```
pratheekshasn, line 22: Why was `shlex.split(command)` removed? Are there going to be any repercussions to removing it?

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,312 @@
...
+        git_obj = GitRepo(
+            local_repo=os.path.basename(config.kernel_src_dir),
+            upstream_repo_url=config.stable_rt_remote,
+            upstream_branch=f"linux-{kernel_version}.y-rt",
+            local_base_branch=config.target_branch,
+            upstream_repo_name="stable-rt",
+            fork_name="origin",
+            fork_url="https://github.com/ni/linux.git"
```
pratheekshasn, line 148: Please don't use hard coded paths all over the file. You can declare them as constants on top of the module and just use that.

scripts/dev/upstream_merge/build_and_install_kernel.py
```diff
@@ -0,0 +1,312 @@
...
+    if not args.dry_run:
+        status, output = run_command(f"git clone {repo_url} {kernel_src_dir}")
+        if status != 0:
+            raise RuntimeError(f"[ERROR] Kernel repo clone failed: {output}")
+        print(f"[INFO] Cloned kernel repository to: {kernel_src_dir}")
+    else:
+        print(f"[INFO] Would clone kernel repository from {repo_url} to: {kernel_src_dir}")
+
+# === Step 2: Robust Upstream Merge with Conflict Handling ===
+def run_upstream_merge_script(args):
```
pratheekshasn, line 130: You might want to consider renaming this function. You're not running a script as part of your function.

pratheekshasn (changes requested):

scripts/dev/upstream_merge/build_and_install_kernel.py
DIVYANSHIKAUSHHIK: okay, i will remove this check.

scripts/dev/upstream_merge/utils/shell_commands.py
DIVYANSHIKAUSHHIK: It was removed because shlex.split() breaks shell operators (like ||, &&, and |). It treats them as literal text arguments rather than control instructions (that was the issue that I explained you earlier).

As far as I researched there are no negative repercussions. In fact, removing it is necessary for our command chaining to work. By passing the raw string with shell=True, we ensure the shell correctly interprets pipes and conditionals

scripts/dev/upstream_merge/build_and_install_kernel.py
DIVYANSHIKAUSHHIK: okay, updated for next commit.

scripts/dev/upstream_merge/utils/toolchain.py
DIVYANSHIKAUSHHIK: It is a module-level constant used to avoid code duplication.

Since the path is used in multiple functions (e.g., _has_build_script and detect_or_build_cross_compile), defining it at the top level ensures consistency and makes it easier to update the path in one place if it changes.

scripts/dev/upstream_merge/utils/toolchain.py
DIVYANSHIKAUSHHIK: updated in latest commit

scripts/dev/upstream_merge/build_and_install_kernel.py
DIVYANSHIKAUSHHIK: updated in latest commit

shubhanshumt26:
We have already made changes in different PR.
Hence closing this stale PR.

PR #187, opened 2022-10-06
estimate-usage: Use alphabetically latest export

chaitu236:
Use alphabetically latest export instead of chronologically latest so that touching an old folder wouldn't result in wrong export being selected.

Someone seems to have touched 22.0 folder (or created, deleted a file in that folder) in last couple of days which resulted in it being selected as latest export causing pipeline failure (as image names didn't match expectations).

### Testing
`ls -r | head -1` results in latest export being selected.

PR #335, opened 2026-02-03
Support virtualized TPM attachments to qemu VMS, plus refactor.

amstewart:
# Changes

1. Fix a warning from QEMU about using an old syntax for marking a QEMU drive as read-only.
2. Add a new option to the linux start-vm script (`-t`). When asserted, the script will create, initialize, and attach a virtualized TPM2.0 device to the VM using the `swtpm` distro package on the host machine.
3. Refactor the QEMU VM creation process. Instead of using a `build.vms.sh` bash script, use a Makefile - where the logic is easeier to parse.
    * Also, use the `ovmf` recipe in OE-core to build the UEFI firmware, instead of using a static copy from *somewhere*.

# Testing

* [x] Built VMs with the new Makefile and tested with them manually.

# Process

* [ ] Depends on [OE-core #187](https://github.com/ni/openembedded-core/pull/187).
* [ ] Depends on [meta-nilrt #937](https://github.com/ni/meta-nilrt/pull/937).

__Suggested Reviewers:__
* @ni/rtos


amstewart:
This PR will break the AZDO build pipeline after it gets merged, because the VM artifacts will be created in a different location - though they are otherwise the same. I'll put in a change to the component Makefiles to account for this.

amstewart:
Closed. Superseded by #336.

PR #345, opened 2026-04-09
Automate regular x86-64 stable RT kernel Upgrade

jatinjb444:
# Changes

#work item - 3181494
https://dev.azure.com/ni/DevCentral/_workitems/edit/3181494

This PR automates fetching and merging the latest stable‑RT kernel tags into the NI Linux Real‑Time branch. It includes clean repository setup, RT merge, kernel build, deployment, and driver rebuild integration using SSHFS and DKMS. Additionally it also include defconfig regeneration and  PR creation automatically once the script ran sucessfully and also sends an email to notify the user about the result status.

## Justification
The RT kernel upgrade process involves multiple manual steps and is error-prone. This automation streamlines the workflow, improves reliability, and ensures consistent execution across runs.

# Testing

- Verified RT tag detection and merge  
- Successfully built and installed kernel on target  
- Confirmed reboot and kernel update using `uname -r`  
- Validated network functionality post-reboot  
- Verified SSHFS mount, header preparation, and DKMS rebuild  
- Confirmed driver status using `dkms status` 

# Process

__Suggested Reviewers:__
@rajendra-desai-ni 
@Shreejit-03 
@ni/rtos

### DIFF
--- a/scripts/dev/upstream_merge/automation_conf.json
+++ b/scripts/dev/upstream_merge/automation_conf.json
@@ -15,15 +15,13 @@
     "ssh_connection":"",
 
     "kernel_build": {
-    "kernel_src_dir": "nilrt-kernel-build/linux",
-    "arch": "x86_64",
-    "temp_modules_dir": "tmp-glibc/modules",
-    "target_branch": "nilrt/master/6.12",
-    "toolchain_prefix": "/usr/local/oecore-x86_64/sysroots/x86_64-nilrtsdk-linux/usr/bin/x86_64-nilrt-linux/x86_64-nilrt-linux-",
+    "kernel_src_dir": "",
+    "arch": "",
+    "temp_modules_dir": "",
+    "target_branch": "",
+    "toolchain_prefix": "",
     "ssh_target": "",   
     "build_host_ip": "",
-    "pr_enabled": false,
-    "pr_target_branch": "nilrt/master/6.12",
     "work_item_id": ""
     }
 }
--- a/scripts/dev/upstream_merge/automation_conf.json
+++ b/scripts/dev/upstream_merge/automation_conf.json
@@ -20,8 +20,10 @@
     "temp_modules_dir": "tmp-glibc/modules",
     "target_branch": "nilrt/master/6.12",
     "toolchain_prefix": "/usr/local/oecore-x86_64/sysroots/x86_64-nilrtsdk-linux/usr/bin/x86_64-nilrt-linux/x86_64-nilrt-linux-",
-    "target_ip": "10.152.9.10",
-    "target_user": "admin",
-    "build_host_ip": "10.152.9.56"
-}
+    "ssh_target": "",   
+    "build_host_ip": "",
+    "pr_enabled": false,
+    "pr_target_branch": "nilrt/master/6.12",
+    "work_item_id": ""
+    }
 }
--- a/scripts/dev/upstream_merge/automation_conf.json
+++ b/scripts/dev/upstream_merge/automation_conf.json
@@ -12,5 +12,16 @@
     "email_to": "",
     "username": "",
     "build_args":"",
-    "ssh_connection":""
+    "ssh_connection":"",
+
+    "kernel_build": {
+    "kernel_src_dir": "nilrt-kernel-build/linux",
+    "arch": "x86_64",
+    "temp_modules_dir": "tmp-glibc/modules",
+    "target_branch": "nilrt/master/6.12",
+    "toolchain_prefix": "/usr/local/oecore-x86_64/sysroots/x86_64-nilrtsdk-linux/usr/bin/x86_64-nilrt-linux/x86_64-nilrt-linux-",
+    "target_ip": "10.152.9.10",
+    "target_user": "admin",
+    "build_host_ip": "10.152.9.56"
+}
 }
--- a/scripts/dev/upstream_merge/copy_toolchain_from_server.py
+++ b/scripts/dev/upstream_merge/copy_toolchain_from_server.py
@@ -70,7 +70,7 @@ def copy_toolchain_from_nirvana(build_root):
 
     # Idempotent: skip if already present
     if os.path.isdir(toolchain_dst):
-        return 0, f"Toolchain already present at {toolchain_dst}"
+        return 0, f"Toolchain already present at {toolchain_dst}", toolchain_dst
 
     os.makedirs(toolchain_root, exist_ok=True)
 
@@ -82,22 +82,21 @@ def copy_toolchain_from_nirvana(build_root):
         f"Toolchain copied from Nirvana\n"
         f"Version: {latest_version}/{latest_build}\n"
         f"Location: {toolchain_dst}"
-    )
+    ), toolchain_dst
 
 
-def prepare_toolchain_environment(build_root):
+def prepare_toolchain_environment(toolchain_dst):
     """
     Run the NILinuxRT SDK installer (non-interactive) if needed,
     extracting into the local build toolchain directory,
     then source the environment.
     """
 
-    toolchain_path = os.path.join(build_root, "toolchain", "NILinuxRT-x64")
-    if not os.path.isdir(toolchain_path):
-        return 1, f"Toolchain directory not found: {toolchain_path}"
+    if not os.path.isdir(toolchain_dst):
+        return 1, f"Toolchain directory not found: {toolchain_dst}"
 
     env_script = None
-    for root, dirs, files in os.walk(toolchain_path):
+    for root, dirs, files in os.walk(toolchain_dst):
         for f in files:
             if f.startswith("environment-setup"):
                 env_script = os.path.join(root, f)
@@ -107,26 +106,26 @@ def prepare_toolchain_environment(build_root):
 
     # If SDK is not yet extracted, run installer
     if not env_script:
-        installers = [
-            f for f in os.listdir(toolchain_path)
-            if f.endswith(".sh")
-        ]
-
-        if not installers:
+        installer = next(
+            (f for f in os.listdir(toolchain_dst) if f.endswith(".sh")),
+            None
+        )
+        if not installer:
             return 1, "No SDK installer (.sh) found in toolchain directory"
 
-        installer = os.path.join(toolchain_path, installers[0])
+        installer = os.path.join(toolchain_dst, installer)
+
         os.chmod(installer, 0o755)
 
-        # IMPORTANT: force extraction into toolchain_path
+        # IMPORTANT: force extraction into toolchain_dst
         ret = os.system(
-            f'bash "{installer}" -y -d "{toolchain_path}"'
+            f'bash "{installer}" -y -d "{toolchain_dst}"'
         )
         if ret != 0:
             return 1, "Failed to run SDK installer"
 
         # Look again for environment-setup after extraction
-        for root, dirs, files in os.walk(toolchain_path):
+        for root, dirs, files in os.walk(toolchain_dst):
             for f in files:
                 if f.startswith("environment-setup"):
                     env_script = os.path.join(root, f)
@@ -144,7 +143,14 @@ def prepare_toolchain_environment(build_root):
             key, _, value = line.strip().partition("=")
             os.environ[key] = value
 
+# CROSS_COMPILE is expected to be exported by the SDK environment-setup script
+
     if "CROSS_COMPILE" not in os.environ:
-        return 1, "CROSS_COMPILE not set after sourcing toolchain"
+        return 1, (
+            "CROSS_COMPILE not set after sourcing toolchain environment. "
+            "Ensure the SDK environment-setup script exports it correctly."
+        )
+
+    print(f"[TOOLCHAIN] CROSS_COMPILE set to: {os.environ['CROSS_COMPILE']}")
 
     return 0, "Toolchain environment ready"
--- a/scripts/dev/upstream_merge/json_config.py
+++ b/scripts/dev/upstream_merge/json_config.py
@@ -2,6 +2,7 @@
     configuration file."""
 
 import json
+from logging import config
 import os
 
 
@@ -9,8 +10,8 @@ class JsonConfig:
     """Handles loading and validating the automation_conf.json
         configuration file."""
 
-    def __init__(self, automation_conf_path, work_item_id):
-        with open(automation_conf_path, "r", encoding="utf-8") as file:
+    def __init__(self, config_path, work_item_id):
+        with open(config_path, "r", encoding="utf-8") as file:
             config = json.load(file)
         self.work_item_id = work_item_id
         self.nilrt_branch = config.get("nilrt_branch")
@@ -27,3 +28,34 @@ class JsonConfig:
         self.log_level = config.get("log_level")
         self.build_args = config.get("build_args", "")
         self.ssh_connection = config.get("ssh_connection")
+        
+        # Kernel build configuration (optional section)
+        kernel_config = config.get("kernel_build", {})
+        # Expand environment variables in paths
+        self.kernel_src_dir = self._expand_path(kernel_config.get("kernel_src_dir"))
+        self.kernel_target_host = kernel_config.get("target_host")
+        self.kernel_target_user = kernel_config.get("target_user")
+        self.arch = kernel_config.get("arch")
+        self.temp_modules_dir = self._expand_path(kernel_config.get("temp_modules_dir"))
+        self.toolchain_prefix = self._expand_path(kernel_config.get("toolchain_prefix"))
+        self.merge_workdir = self._expand_path(kernel_config.get("merge_workdir"))
+        self.target_branch = kernel_config.get("target_branch")
+        self.stable_rt_remote = kernel_config.get("stable_rt_remote")
+        self.nilrt_root = self._expand_path(kernel_config.get("nilrt_root"))
+        self.required_packages = kernel_config.get("required_packages", [])
+        self.make_jobs = kernel_config.get("make_jobs", "$(nproc)")
+        self.kernel_config = kernel_config.get("kernel_config", "defconfig")
+        self.repo_url = kernel_config.get("repo_url")
+        self.target_name = kernel_config.get("target_name")
+        self.target_ip = kernel_config.get("target_ip")
+        self.target_user = kernel_config.get("target_user", "admin")
+        self.build_host_ip = kernel_config.get("build_host_ip")
+        self.ssh_options = kernel_config.get("ssh_options", "-o StrictHostKeyChecking=no")
+        self.build_host_user = kernel_config.get("build_host_user", os.getenv("USER", "builduser"))
+
+    
+    def _expand_path(self, path):
+        """Expand environment variables and user home directory in paths."""
+        if path:
+            return os.path.expandvars(os.path.expanduser(path))
+        return path
--- a/scripts/dev/upstream_merge/json_config.py
+++ b/scripts/dev/upstream_merge/json_config.py
@@ -2,7 +2,6 @@
     configuration file."""
 
 import json
-from logging import config
 import os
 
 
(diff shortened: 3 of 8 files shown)


Shreejit-03:
Please run flake8 ( python3 -m flake8 <file>) on all the python files and fix all the issues we can. Some of them could be fixed by running black (python3 -m black <file>).

scripts/dev/upstream_merge/json_config.py
```diff
@@ -9,8 +9,8 @@ class JsonConfig:
     """Handles loading and validating the automation_conf.json
         configuration file."""
 
-    def __init__(self, automation_conf_path, work_item_id):
-        with open(automation_conf_path, "r", encoding="utf-8") as file:
+    def __init__(self, config_path, work_item_id):
```
Shreejit-03, line 12: Any particular reason to rename it?

scripts/dev/upstream_merge/json_config.py
```diff
@@ -27,3 +27,43 @@ def __init__(self, automation_conf_path, work_item_id):
         self.log_level = config.get("log_level")
         self.build_args = config.get("build_args", "")
         self.ssh_connection = config.get("ssh_connection")
+        # Kernel build configuration (optional section)
+        kernel_config = config.get("kernel_build", {})
+        # Expand environment variables in paths
```
Shreejit-03, line 31: I don't see the need for using environment variables

scripts/dev/upstream_merge/json_config.py
```diff
@@ -27,3 +27,43 @@ def __init__(self, automation_conf_path, work_item_id):
...
+            "ssh_options", "-o StrictHostKeyChecking=no")
+        self.build_host_user = kernel_config.get(
+            "build_host_user", os.getenv("USER", "builduser"))
+        self.pr_enabled = config.get("pr_enabled", False)
+        self.pr_target_branch = config.get(
+            "pr_target_branch", self.target_branch)
+        self.work_item_id = config.get("work_item_id", "")
+        self.ssh_target = kernel_config.get("ssh_target")
+
+    def _expand_path(self, path):
```
Shreejit-03: If we aren't making use of any environment variables, then this can be removed.

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -0,0 +1,586 @@
...
+        )
+        if ret == 0:
+            print("[INFO] SSH is available on target")
+            return 0
+        time.sleep(5)
+
+    return 1
+
+
+def run_cmd(cmd):
```
Shreejit-03, line 82: Is run_cmd any different from [run_command](https://github.com/ni/nilrt/blob/140e6f185e94417cff69e5ade87cff6843f6dd19/scripts/dev/upstream_merge/utils/shell_commands.py#L11) in its usage and implementation? It would be better to use [run_command](https://github.com/ni/nilrt/blob/140e6f185e94417cff69e5ade87cff6843f6dd19/scripts/dev/upstream_merge/utils/shell_commands.py#L11) .

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -0,0 +1,586 @@
...
+        git_merge_abort()
+    except Exception:
+        pass
+
+    git_reset(hard=True)
+    git_clean(force=True, directories=True, ignored_files=True)
+
+    git_fetch("origin")
+    git_checkout(config.target_branch, create=True, force_checkout=True)
+    git_reset(hard=True, target=f"origin/{config.target_branch}")
```
Shreejit-03: As the ni/linux repo will be freshly cloned, I don't really see the need for git_merge_abort(), git_reset(), git_clean().

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -0,0 +1,586 @@
...
+
+
+# --------------------
+# Kernel repo handling
+# --------------------
+def clone_kernel_repository():
+    print("[INFO] Cloning kernel repository")
+
+    repo = os.path.abspath(config.kernel_src_dir)
+    config.kernel_src_dir = repo
```
Shreejit-03: This can be handled in [json_config.py](https://github.com/ni/nilrt/blob/140e6f185e94417cff69e5ade87cff6843f6dd19/scripts/dev/upstream_merge/json_config.py#L18).

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -0,0 +1,586 @@
...
+
+    git_reset(hard=True)
+    git_clean(force=True, directories=True, ignored_files=True)
+
+    git_fetch("origin")
+    git_checkout(config.target_branch, create=True, force_checkout=True)
+    git_reset(hard=True, target=f"origin/{config.target_branch}")
+
+    _, remotes = git_remote()
+    if "stable-rt" not in remotes:
```
Shreejit-03: Is there a reason to not use kernel_repo.[add_remote()](https://github.com/ni/nilrt/blob/140e6f185e94417cff69e5ade87cff6843f6dd19/scripts/dev/upstream_merge/utils/git_repo.py#L125)?

scripts/dev/upstream_merge/json_config.py
jatinjb444: It is to make sure that the json config file is a generic config loader file, and not tied to work with automation_conf only.

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Yes—run_cmd differs intentionally. run_command relies on shlex.split, which breaks SSH commands that include shell operators (||, &&, pipes, quoted remote commands). The rebuild/install steps depend on shell semantics, so run_cmd executes via the shell to preserve existing behavior.

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: It was just added for an extra check, but yes it can be removed if we are assuming it will work for cloning freshly every time .

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Sure , i have handled it in the json_config.py

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: so since we already have a GitRepo instance here, I’ve switched to using kernel_repo.add_remote() to keep git operations encapsulated and consistent

scripts/dev/upstream_merge/json_config.py
jatinjb444: Yeah , i have changed it to standardize the overall json config file and use the paths explicitly.

scripts/dev/upstream_merge/json_config.py
jatinjb444: Agreed — since we’re not relying on environment variable expansion right now, I can remove _expand_path and keep the paths explicit.

scripts/dev/upstream_merge/utils/build.py
```diff
@@ -122,6 +127,104 @@ def build_core_images(args=""):
...
+    Returns True if a commit was created, False otherwise.
+    """
+    print("[US4] Regenerating nati_x86_64_defconfig")
+
+    original_cwd = os.getcwd()
+    os.chdir(kernel_src_dir)
+
+    cmds = [
+        "make mrproper",
+        "make nati_x86_64_defconfig",
```
rajendra-desai-ni: You have already defined a global variable KERNEL_DEFCONFIG = "nati_x86_64_defconfig", can't we just reference the variable here (and other places) instead of hardcoding again?

scripts/dev/upstream_merge/utils/build.py
```diff
@@ -122,6 +127,104 @@ def build_core_images(args=""):
...
+    # Ensure ARCH is correct
+    os.environ["ARCH"] = "x86_64"
+
+    # Step 1: Configure kernel
+    ret = os.system(f"make {KERNEL_DEFCONFIG}")
+    if ret != 0:
+        return 1, "Kernel defconfig failed"
+
+    # Step 2: Build kernel and modules
+    ret = os.system(f"make -j{jobs} bzImage modules")
```
rajendra-desai-ni, line 160: Can't we just use `nproc` here instead of reading from config using 'jobs' variable?

scripts/dev/upstream_merge/automation_conf.json
```diff
@@ -12,5 +12,18 @@
...
     "username": "",
     "build_args":"",
-    "ssh_connection":""
+    "ssh_connection":"",
+
+    "kernel_build": {
+    "kernel_src_dir": "nilrt-kernel-build/linux",
+    "arch": "x86_64",
+    "temp_modules_dir": "tmp-glibc/modules",
+    "target_branch": "nilrt/master/6.12",
```
rajendra-desai-ni: It would be better to move all these variable values into pipeline env variables instead of hardcoding them here, similar to how it was done for oe merge automation.

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -0,0 +1,480 @@
...
+REPO_URL = "https://github.com/ni/linux.git"
+STABLE_RT_REMOTE = (
+    "https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-stable-rt.git"
+)
+
+
+# --------------------
+# Kernel build config (x86_64)
+# --------------------
+KERNEL_DEFCONFIG = "nati_x86_64_defconfig"
```
rajendra-desai-ni: Is this variable redeclaration needed as you are already importing build(.py) code which has this and below variables set?

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -0,0 +1,480 @@
...
+            merge_report=merge_report,
+            email_log_level=0,
+            skip_push_and_pr=True,
+        )
+        return
+
+    # --------------------
+    # Success case
+    # --------------------
+    print("[INFO] RT merge successful")
```
rajendra-desai-ni, line 254: Have you tested fail case works as well?

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -0,0 +1,480 @@
...
+    print("[INFO] Kernel install issued reboot request")
+
+    time.sleep(20)
+
+    print("[INFO] Waiting for target to come back after kernel reboot")
+    if wait_for_ssh(config.ssh_target, timeout=300) != 0:
+        print("[ERROR] Target did not come back after kernel reboot")
+        return
+
+    print("[INFO] Target is back online, starting rebuild drivers")
```
rajendra-desai-ni: You could move 'rebuild drivers' steps into a new function since the existing function 'run_upstream_merge_script' mainly means to just do the upstream merge and should not include testing or validation part in it.

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -0,0 +1,480 @@
...
+    if rc != 0:
+        print("[ERROR] STEP 3 failed")
+        return
+
+    # STEP 4: DKMS rebuild
+    rc = dkms_autoinstall(config, run_cmd)
+    if rc != 0:
+        print("[ERROR] STEP 4 failed")
+        return
+    # PR creation (FINAL STEP ONLY) ----
```
rajendra-desai-ni: You could move 'publish PR' steps into a new function since the existing function 'run_upstream_merge_script' mainly means to just do the upstream merge and should not include PR creation part in it.

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -0,0 +1,480 @@
...
+        merge_report=merge_report,
+        email_log_level=0,
+        skip_push_and_pr=args.skip_push_and_pr,
+    )
+    return
+
+
+def install_kernel_to_target(kernel_src_dir):
+    """
+    Install x86_64 kernel and modules to an NILRT target (cRIO).
```
rajendra-desai-ni: You don't have to specify as `cRIO` since the scripts should also be able to test a PXI controller.

scripts/dev/upstream_merge/copy_toolchain_from_server.py
rajendra-desai-ni, line 1: You are just copying toolchain SDK from the server and not actually building one. Better to rename the file to `copy_toolchain_from_server.py` or something.

scripts/dev/upstream_merge/copy_toolchain_from_server.py
```diff
@@ -0,0 +1,150 @@
...
+            return 1, "installer ran , env-setup script not found"
+
+    # Source environment into Python process
+    cmd = f'bash -c "source {env_script} >/dev/null && env"'
+    with os.popen(cmd) as proc:
+        for line in proc:
+            key, _, value = line.strip().partition("=")
+            os.environ[key] = value
+
+    if "CROSS_COMPILE" not in os.environ:
```
rajendra-desai-ni, line 148: Where are you setting `CROSS_COMPILE` variable?

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -0,0 +1,480 @@
...
+        return 1, "depmod failed on target (ABORTING reboot)"
+
+    # Reboot target ONLY after everything succeeded
+    rc = os.system(f"ssh {ssh_opts} {ssh_target} reboot")
+    if rc != 0:
+        return 1, "FAILED to reboot target after install"
+
+    return 0, (
+        f"Kernel and modules installed successfully\n"
+        f"Kernel version: {kernel_version}"
```
rajendra-desai-ni, line 491: You need to also test if the `kernel_version` matches `uname -r` on the target after installing the newer kernel. Looks like you are not doing that anywhere, which is one of the basic test to confirm if the upstream merge actually worked.

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -0,0 +1,480 @@
...
+    # CRITICAL: rebuild module dependency indexes
+    rc = os.system(
+        f"ssh {ssh_opts} {ssh_target} "
+        f'"depmod -a {kernel_version}"'
+    )
+    if rc != 0:
+        return 1, "depmod failed on target (ABORTING reboot)"
+
+    # Reboot target ONLY after everything succeeded
+    rc = os.system(f"ssh {ssh_opts} {ssh_target} reboot")
```
rajendra-desai-ni, line 485: Can you also include [this](https://github.com/ni/nilrt/blob/140e6f185e94417cff69e5ade87cff6843f6dd19/docs/README.kernel.md?plain=1#L285) step in the code before rebooting so that the target can be accessed in safemode in case when there is a bad kernel installed?

rajendra-desai-ni (changes requested):
- [This](https://github.com/ni/nilrt/pull/345/changes/f075a8bf80fa1d32d100d455e601e1ff3ef7f0a6) commit doesn't have a proper title.
- Remove "Finalized" word from the PR title.
- Add AB work item in the PR description.
- You don't have to add 'Updates' section in the PR description. All the updates should go into their respective work items.
- Try to follow 'Summary', 'Justification', and 'Testing' section in the PR description which is cleaner compared to existing sections.

scripts/dev/upstream_merge/utils/build.py
jatinjb444: We will lose the ability to limit jobs , instead I have updated it to read it from config , if not available it will directly use the nproc (safer option)

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Yes , it works and notifies the user via email

scripts/dev/upstream_merge/copy_toolchain_from_server.py
jatinjb444: CROSS_COMPILE is set via the SDK environment-setup script during sourcing. Added explicit validation and logging to ensure it is correctly exported and to improve clarity.

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Yeah,  Added kernel version validation using uname -r after reboot and SSH restore, before proceeding with driver rebuild.

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Done

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Moved PR creation logic into a separate function to maintain separation of concerns and keep upstream merge workflow focused on core steps.

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Added fw_setenv bootdelay configuration before reboot to allow safe mode access in case of kernel issues on the target.

scripts/dev/upstream_merge/copy_toolchain_from_server.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def prepare_toolchain_environment(build_root):
+    """
+    Run the NILinuxRT SDK installer (non-interactive) if needed,
+    extracting into the local build toolchain directory,
+    then source the environment.
+    """
+
+    toolchain_path = os.path.join(build_root, "toolchain", "NILinuxRT-x64")
```
rajendra-desai-ni: Can't this variable be replaced with `toolchain_dst` that you have defined in `copy_toolchain_from_nirvana()` function by making it global or replace `toolchain_dst` with this variable? Why do we need to create different variables with same values?

scripts/dev/upstream_merge/copy_toolchain_from_server.py
```diff
@@ -0,0 +1,157 @@
...
+    if not env_script:
+        installers = [
+            f for f in os.listdir(toolchain_path)
+            if f.endswith(".sh")
+        ]
+
+        if not installers:
+            return 1, "No SDK installer (.sh) found in toolchain directory"
+
+        installer = os.path.join(toolchain_path, installers[0])
```
rajendra-desai-ni: When you are always extracting `installers` value at 0th index, why do you need to even make `installers` as a list? Can't it just be a string instead?

rajendra-desai-ni (approved):

scripts/dev/upstream_merge/copy_toolchain_from_server.py
jatinjb444: Sure , i have reused it from [copy_toolchain_from_nirvana()]

scripts/dev/upstream_merge/copy_toolchain_from_server.py
jatinjb444: Updated installer selection logic to directly fetch the first matching .sh file without creating an intermediate list, improving clarity and efficiency

chaitu236 (changes requested):
Convention for commit message header is "recipe/subsystem: commit header". This lets us know just by looking at a commit header which files/subsystem/recipe a commit is touching.

Please edit the commit messages to follow this pattern.

E.g., like
`upstream_merge: Refactor build logic and clean up scripts`

Also we try to keep commit header not too long (50 chars is ideal but we regularly go beyond that if we can't fit in 50 chars).
Following commit will become too long though, try to see if we can shorten it.

`upstream_merge: Merge stable RT patches, build kernel, and validate target deployment` this will become very long.

PR #347, opened 2026-04-21
docker: add macvlan network setup script

Shreejit-03:
# Changes
Add `setup-nilrt-network.sh` creates a macvlan Docker/podman network (nilrt-net) giving each NILRT container its own
IP address on the physical LAN. This allows containers to be discovered and accessed directly from remote hosts.

## Key features:
- Auto-detects the host's network interface, subnet, and gateway from the routing table; all values can be overridden via CLI options.
- Creates a macvlan shim interface (nilrt-shim) on the host to work around the kernel limitation where macvlan parent interfaces cannot communicate with their child interfaces, enabling host-to-container connectivity.
- Enables promiscuous mode on the parent NIC so frames addressed to container MAC addresses are not dropped (required in VM environments such as VirtualBox, VMware, and Hyper-V).
- Supports both Docker and podman (rootful) runtimes.
- Includes --dry-run mode to preview commands without executing.
- Idempotent: detects existing networks and shim interfaces and skips re-creation.

# Justification
[AB#3739210](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3739210).

# Testing

- Launched `nilrt-slim-container:11.5-slim` and `nilrt-runmode-container:11.5` on the `nilrt-net` network. Verified each container received a unique IP on the physical subnet.
- Confirmed host-to-container connectivity & remote connectivity: `ping <container-ip>`.
- Tested on a VM with promiscuous mode enabled. Verified containers were reachable from both the VM host and remote machines on the network.
- Tested --dry-run mode: verified commands are printed but not executed.
- Tested idempotent re-runs: re-running the script when the network and shim already exist skips re-creation without errors.
- Ran with --runtime podman (rootful). Verified network creation and container launch succeed and recieved a unique IP which was accessible.(Note: podman requires sudo for macvlan — rootless podman cannot create macvlan interfaces due to missing CAP_NET_ADMIN in the user namespace.)

### DIFF
--- a/docker/setup-nilrt-network.sh
+++ b/docker/setup-nilrt-network.sh
@@ -1,0 +1,318 @@
+#!/bin/bash
+# Create a macvlan network for NILRT containers so they are accessible
+# from remote hosts on the LAN.
+#
+# Macvlan gives each container its own IP address on the physical network,
+# making it behave like a standalone machine. Remote hosts can reach the
+# container directly without port forwarding.
+#
+# Prerequisites:
+#   - Docker or podman must be installed.
+#   - The host's physical network interface must be up.
+#   - An IP range on the LAN must be reserved for containers (to avoid
+#     DHCP conflicts). Coordinate with your network administrator.
+#
+# Usage:
+#   ./setup-nilrt-network.sh [OPTIONS]
+#
+# After setup, run containers on this network:
+#   docker run -it --network=nilrt-net nilrt-slim-container:11.5-slim
+#   docker run -it --network=nilrt-net nilrt-runmode-container:11.5
+#
+#   podman run -it --network=nilrt-net nilrt-slim-container:11.5-slim
+#
+# Examples:
+#   ./setup-nilrt-network.sh
+#   ./setup-nilrt-network.sh -i eth0 -s 192.0.2.0/24 -g 192.0.2.1 -r 192.0.2.64/26
+#   ./setup-nilrt-network.sh --runtime podman
+
+set -euo pipefail
+
+# Defaults — adjust to match your LAN.
+PARENT_IFACE=""
+SUBNET=""
+GATEWAY=""
+IP_RANGE=""
+NETWORK_NAME="nilrt-net"
+RUNTIME="docker"
+DRY_RUN=""
+SHIM_IFACE="nilrt-shim"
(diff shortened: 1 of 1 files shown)


docker/setup-nilrt-network.sh
```diff
@@ -0,0 +1,271 @@
...
+else:
+    print(net.broadcast_address - 1)" "$1"
+}
+
+# Convert a host/prefix to its network address (e.g. 192.0.2.100/24 -> 192.0.2.0/24).
+network_address() {
+	python3 -c "
+import ipaddress, sys
+print(ipaddress.ip_network(sys.argv[1], strict=False))" "$1"
+}
```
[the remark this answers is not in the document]
rajendra-desai-ni: `python3` installation check can be ignored since most of the Linux host systems have it pre-installed.

docker/setup-nilrt-network.sh
```diff
@@ -0,0 +1,271 @@
...
+
+# Auto-detect the default route interface, subnet, and gateway.
+auto_detect_network() {
+	if [[ -z "$PARENT_IFACE" ]]; then
+		PARENT_IFACE=$(ip -o route get 8.8.8.8 2>/dev/null | sed -n 's/.*dev \([^ ]*\).*/\1/p')
```
chaitu236, line 116: Why not `ip route show default`? It seems odd to reference a specific IP.

chaitu236 (commented):
I didn't get to review the full script and may not be able to find time to do it anytime soon. I'll leave this to @rajendra-desai-ni to review and merge. Other RTOS members are also welcome to review.

docker/setup-nilrt-network.sh
Shreejit-03: Using `ip route show default` only tells us what the system has configured as the default gateway, but it doesn’t necessarily reflect how traffic is actually routed in all cases. For example:

- With multiple interfaces (Ethernet, Wi-Fi, VPN), there may be multiple default routes or different metrics.
- With policy-based routing or multiple routing tables, the effective path can differ from the “default” entry.
- In some environments, there may not even be a traditional default route, but specific routes still exist.

rajendra-desai-ni (approved):

PR #351, opened 2026-05-25
docker: add nilrt-ctr.sh container management CLI

Shreejit-03:
# Changes
Add a Docker Compose configuration and `nilrt-ctr.sh` CLI for managing NILRT containers locally.

1. `docker-compose.yml` — Defines two services (`nilrt` runmode and `nilrt-slim`) with shared configuration (privileged, healthcheck on niauth_daemon, macvlan networking via the external nilrt-net network). Supports scaling to run multiple container instances.
2. `nilrt-ctr.sh` — A Bash CLI for NILRT-specific container operations. Non-NILRT operations are intentionally left to native `docker`/`docker compose`:
    - `status` / `info` — concise target summary (name, ID, image, state, started, hostname, IP).
    - `set-feed` / `feed` — configure opkg package feeds in `/etc/opkg/base-feeds.conf`.
    - `install` — install packages via opkg with lock-wait retry; optional `--feed YYYYQN` flag.
    - `remove` / `uninstall` — remove packages via opkg.
    - `change-hostname` / `hostname` — set container hostname.
    - `shell` / `ssh` — open an interactive shell.
    - `scale` — scale service instances via docker compose.
    - Target resolution by container name, ID prefix, or numeric index.

Requires the nilrt-net macvlan network to be created beforehand (via `setup-nilrt-network.sh`).

# Justification
[AB#3201995](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3201995).

# Testing
All commands were manually tested against running NILRT containers:

| Command | Test |
|---------|------|
| `status \| info <target>` | Displays name, ID, image, state, started time, hostname, and IP |
| `set-feed \| feed <target\|all> <YYYYQN>` | Writes correct feed entries to `/etc/opkg/base-feeds.conf`; verified with `cat` inside container; re-running replaces entries without duplicating; rejects invalid YYYYQN format |
| `install <target> [--feed YYYYQN] <pkg...>` | Sets feed (if given), runs `opkg update`, installs packages; verified with `opkg list-installed` |
| `remove \| uninstall <target> <pkg...>` | Removes packages; confirmed package no longer listed; defers to opkg's dependency resolution for packages with dependents |
| `change-hostname \| hostname <target> <name>` | Sets hostname inside container; verified with `hostname` |
| `shell \| ssh <target>` | Opens interactive bash session |
| `scale <service> <n>` | Scaled `nilrt`/`nilrt-slim`; new containers joined macvlan network with unique IPs |

Additionally verified:
- Target resolution works by container name, ID prefix, and numeric index
- Error handling for invalid targets, missing arguments, unknown commands, and out-of-range indices
- `--feed` flag on `install` sets feeds before installing
- Image version auto-detected from local `nilrt-runmode-container` tags for `scale`

### DIFF
--- a/docker/nilrt-ctr.sh
+++ b/docker/nilrt-ctr.sh
@@ -1,0 +1,326 @@
+#!/bin/bash
+# nilrt-ctr.sh — Management CLI for NILRT containers.
+#
+# Provides discovery, configuration, software management, and monitoring
+# for NILRT containers managed by docker-compose.yml.
+#
+# Usage:
+#   bash docker/nilrt-ctr.sh <command> [arguments...]
+#
+# Commands:
+#   status|info <target>                       Show target summary
+#   set-feed|feed <target|all> <YYYYQN>        Set package feed
+#   install <target> [--feed YYYYQN] <pkg...>  Install packages on a target
+#   remove|uninstall <target> <pkg...>         Remove packages from a target
+#   change-hostname <target> <hostname>        Set target hostname
+#   shell|ssh <target>                         Open an interactive shell on a target
+#   scale <service> <n>                        Scale a service to n instances
+#
+# Targets can be specified by container name, container ID (prefix), or
+# index from managed container list (e.g. "1", "2").
+#
+# Examples:
+#   bash docker/nilrt-ctr.sh status nilrt-slim-1
+#   bash docker/nilrt-ctr.sh set-feed all 2026Q2
+#   bash docker/nilrt-ctr.sh install nilrt-1 --feed 2026Q2 ni-labview-realtime
+#   bash docker/nilrt-ctr.sh change-hostname nilrt-1 cRIO-test
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+COMPOSE_FILE="${SCRIPT_DIR}/docker-compose.yml"
+DOCKER_CMD="docker"
+LABEL_FILTER="label=nilrt.managed=true"
+
+# Auto-detect NILRT image version from local docker images (used by 'scale').
+if [[ -z "${NILRT_VERSION:-}" ]]; then
+    NILRT_VERSION=$(docker images --format '{{.Tag}}' nilrt-runmode-container 2>/dev/null \
+        | grep -v '^<none>$' | sort -V | tail -n1)
+    [[ -z "$NILRT_VERSION" ]] && NILRT_VERSION="latest"
+fi
+export NILRT_VERSION
+
+# Colors
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+NC='\033[0m'
+
+usage() {
+    cat <<'EOF'
+Usage: bash docker/nilrt-ctr.sh <command> [arguments...]
+
+  Management CLI for NILRT containers.
+
+Commands:
+  status|info <target>                       Show target summary
+  set-feed|feed <target|all> <YYYYQN>        Set package feed
+  install <target> [--feed YYYYQN] <pkg...>  Install packages
+  remove|uninstall <target> <pkg...>         Remove packages
+  change-hostname <target> <hostname>        Set target hostname
+  shell|ssh <target>                         Interactive shell
+  scale <service> <n>                        Scale service instances
+
+Targets: container name, ID prefix, or managed index.
+
+Use native docker for non-NILRT-specific operations, for example:
+    docker ps -a --filter label=nilrt.managed=true
+    docker logs <container>
+    docker exec -it <container> /bin/bash
+    docker compose -f docker/docker-compose.yml up -d --scale nilrt=3
+EOF
+    exit "${1:-0}"
+}
+
+err() { echo -e "${RED}ERROR:${NC} $*" >&2; exit 1; }
+warn() { echo -e "${YELLOW}WARNING:${NC} $*" >&2; }
+info() { echo -e "${BLUE}==>${NC} $*"; }
+
+# Wait for opkg lock to be released inside a container, then run opkg command.
+# Retries the opkg command directly since wrapping with flock conflicts with
+# opkg's internal lock acquisition (same file, different fd).
+run_opkg() {
+    local cid="$1"; shift
+    local retries=30
+    local output rc target_name
+    target_name=$($DOCKER_CMD inspect --format '{{.Name}}' "$cid" 2>/dev/null | sed 's|^/||')
+    target_name="${target_name:-${cid:0:12}}"
+    while [[ $retries -gt 0 ]]; do
+        output=$($DOCKER_CMD exec "$cid" opkg "$@" 2>&1)
+        rc=$?
+        if [[ $rc -eq 0 ]]; then
+            [[ -n "$output" ]] && echo "$output"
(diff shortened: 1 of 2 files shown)


docker/nilrt-ctr.sh
rajendra-desai-ni, line 1: I assume this script mostly uses existing `docker` commands under the hood and acts as a docker wrapper for most of the functions. Though having a script is a cleaner way to run docker commands that can also execute nilrt container specific ones, I still feel we shouldn't be wrapping unneeded options which can be directly run with `docker`. The reason is that Docker has a ton of options that can be used to operate on containers, but we are not wrapping all those options in the script. This would force users to still use docker directly for those options. Also, having wrappers unnecessarily can cause them to break when the options get updated/changed in the docker.

docker/nilrt-ctr.sh
```diff
@@ -0,0 +1,493 @@
...
+}
+
+# Get the container name for a given ID
+container_name() {
+    $DOCKER_CMD inspect --format '{{.Name}}' "$1" 2>/dev/null | sed 's|^/||'
+}
+
+# ---- Commands ----
+
+cmd_discover() {
```
rajendra-desai-ni: This function just acts as a status option and doesn't actually make any container changes, we can use the normal `docker` with `grep` to filter nilrt specific ones. So, I would suggest removing this function. But you can let me know why it would still be useful.

docker/nilrt-ctr.sh
```diff
@@ -0,0 +1,493 @@
...
+
+        printf "%-4s %-12s %-30s %-35s ${state_color}%-12s${NC} %-18s\n" \
+            "$idx" "${cid:0:12}" "$name" "$image" "$state" "${ip:-N/A}"
+        ((idx++))
+    done
+    echo ""
+    echo -e "Total: $((idx - 1)) target(s)"
+}
+
+cmd_status() {
```
rajendra-desai-ni: I am fine with this function but try to simplify it if possible. We can skip showing details such as "System Information" (because `docker ps` would give that), os-release info (because image name already has a release version), ni-rt.ini info, FirmwareVersion, kernel version (because it doesn't make sense), Installed NI packages and running services (because list will be too long and not so useful).

docker/nilrt-ctr.sh
```diff
@@ -0,0 +1,493 @@
...
+        lv_line=\"src/gz ${lv_feed} \${feeds_uri}/${feed}/${lv_feed}\"
+        sed -i '/^src\/gz ni-software /d' /etc/opkg/base-feeds.conf 2>/dev/null || true
+        sed -i '/^src\/gz ni-lv[0-9]/d' /etc/opkg/base-feeds.conf 2>/dev/null || true
+        echo \"\${ni_line}\" >> /etc/opkg/base-feeds.conf
+        echo \"\${lv_line}\" >> /etc/opkg/base-feeds.conf
+    "
+}
+
+# Get all managed container IDs
+get_managed_containers() {
```
rajendra-desai-ni: This acts just as a status function without much of usefulness, so it can be removed. Users could just run `docker` command directly instead.

docker/nilrt-ctr.sh
```diff
@@ -0,0 +1,493 @@
...
+    $DOCKER_CMD ps -a --filter "$LABEL_FILTER" --format '{{.ID}}' 2>/dev/null
+}
+
+# Get managed containers as a table
+get_managed_table() {
```
rajendra-desai-ni: This acts just as a status function without much of usefulness, so it can be removed. Users could just run `docker` command directly instead and can be replaced by `cmd_status`.

docker/nilrt-ctr.sh
```diff
@@ -0,0 +1,493 @@
...
+    if [[ -n "$id" ]]; then
+        echo "$id"
+        return
+    fi
+
+    err "Target '${target}' not found. Use 'discover' to list targets."
+}
+
+# Get the container name for a given ID
+container_name() {
```
rajendra-desai-ni: We can directly use docker commands so this function could be removed.

docker/nilrt-ctr.sh
```diff
@@ -0,0 +1,493 @@
...
+    local name
+    name=$(container_name "$cid")
+
+    info "Removing from ${name}: $*"
+    wait_for_opkg_lock "$cid"
+    $DOCKER_CMD exec "$cid" opkg remove "$@"
+    info "Removal complete."
+}
+
+cmd_update() {
```
rajendra-desai-ni: I don't think this function is necessary as cmd_install() already does opkg update.

docker/nilrt-ctr.sh
```diff
@@ -0,0 +1,493 @@
...
+        fi
+        info "Updating ${name}..."
+        wait_for_opkg_lock "$cid"
+        $DOCKER_CMD exec "$cid" opkg update
+        $DOCKER_CMD exec "$cid" opkg upgrade
+        info "${name} updated."
+    done
+}
+
+cmd_rename() {
```
rajendra-desai-ni: You could rename this function to `cmd_change_hostname()` or something to look more clearer.

docker/nilrt-ctr.sh
```diff
@@ -0,0 +1,493 @@
...
+    local cid
+    cid=$(resolve_target "$1")
+    local name
+    name=$(container_name "$cid")
+
+    info "Opening shell on ${name}..."
+    $DOCKER_CMD exec -it "$cid" /bin/bash
+}
+
+cmd_scale() {
```
rajendra-desai-ni, line 299: What is the use case of this function?

docker/nilrt-ctr.sh
```diff
@@ -0,0 +1,493 @@
...
+    if ! [[ "$count" =~ ^[0-9]+$ ]]; then
+        err "Count must be a positive integer."
+    fi
+
+    info "Scaling ${service} to ${count} instance(s)..."
+    "${COMPOSE_CMD[@]}" up -d --scale "${service}=${count}" "$service"
+    info "${service} scaled to ${count}."
+}
+
+cmd_log() {
```
rajendra-desai-ni: We could remove this function as it could be run directly using `docker` command.

docker/nilrt-ctr.sh
```diff
@@ -0,0 +1,493 @@
...
+    [[ $# -lt 1 ]] && err "Usage: nilrt-ctr.sh log <target> [lines]"
+    local cid
+    cid=$(resolve_target "$1")
+    local lines="${2:-50}"
+
+    $DOCKER_CMD logs --tail "$lines" "$cid"
+}
+
+cmd_exec() {
```
rajendra-desai-ni: I don't think there is a great use case of this function, and it can also be run directly with `docker` so we could also remove this.

docker/nilrt-ctr.sh
```diff
@@ -0,0 +1,493 @@
...
+
+    $DOCKER_CMD exec "$cid" "$@"
+}
+
+# ---- Main ----
+
+[[ $# -eq 0 ]] && usage
+
+case "$1" in
+    discover|list|find)   shift; cmd_discover "$@" ;;
```
rajendra-desai-ni: Try to keep function names self-explanatory instead of having one word suffix.

rajendra-desai-ni (changes requested):

rajendra-desai-ni (approved):

PR #352, opened 2026-06-02
upstream_merge: Add automated SSH setup for target-build

jatinjb444:
# Changes
Added automated password-less SSH setup from target to build machine
Generates SSH keys on target and installs them safely on build host
Avoids duplicate key insertion with idempotent handling
Removes dependency on manual SSH configuration for script execution

# Testing
Verified SSH setup runs without manual intervention after initial run
Confirmed target can SSH into build machine without password
Executed full script flow to ensure no failures in SSH stage
Validated no duplicate keys are added across multiple runs

# Process
Implemented SSH automation without modifying existing core workflow
Ensured compatibility with dynamic build user handling
Maintained pipeline-safe logic without interactive input
Followed existing project structure and commit guidelines

__Suggested Reviewers:__
@rajendra-desai-ni 
@ni/rtos

### DIFF
--- a/scripts/dev/upstream_merge/automation_conf.json
+++ b/scripts/dev/upstream_merge/automation_conf.json
@@ -22,6 +22,8 @@
     "toolchain_prefix": "",
     "ssh_target": "",   
     "build_host_ip": "",
+    "build_user": "",
     "work_item_id": ""
     }
 }
+
--- a/scripts/dev/upstream_merge/json_config.py
+++ b/scripts/dev/upstream_merge/json_config.py
@@ -9,8 +9,8 @@ class JsonConfig:
     """Handles loading and validating the automation_conf.json
         configuration file."""
 
-    def __init__(self, config_path, work_item_id):
-        with open(config_path, "r", encoding="utf-8") as file:
+    def __init__(self, automation_conf_path, work_item_id):
+        with open(automation_conf_path, "r", encoding="utf-8") as file:
             config = json.load(file)
         self.nilrt_branch = config.get("nilrt_branch")
         self.meta_nilrt_branch = config.get("meta_nilrt_branch")
@@ -50,6 +50,7 @@ class JsonConfig:
         self.build_host_user = kernel_config.get(
             "build_host_user", os.getenv("USER", "builduser"))
         self.pr_enabled = config.get("pr_enabled", False)
+        self.build_user = kernel_config.get("build_user") or self.build_host_user  
         self.pr_target_branch = config.get(
             "pr_target_branch", self.target_branch)
         self.work_item_id = work_item_id or config.get("work_item_id", "")
--- a/scripts/dev/upstream_merge/kernel_build_and_test.py
+++ b/scripts/dev/upstream_merge/kernel_build_and_test.py
@@ -4,6 +4,7 @@ import sys
 import re
 import argparse
 import time
+import getpass
 from log_and_email_utils import setup_logging, write_log_and_send_email
 from utils import build
 from utils.git_commands import (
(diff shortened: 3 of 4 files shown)


scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -491,6 +575,20 @@ def install_kernel_to_target(kernel_src_dir):
...
 
+
+def setup_environment():
+    print("[SETUP] Setting up environment...")
+    # --------------------
+    # Git config (ONLY if missing)
+    # --------------------
+    print("[SETUP] Configuring git identity...")
+    if os.system("git config --global user.name >/dev/null 2>&1") != 0:
+        os.system('git config --global user.name "Jatin Bharti"')
```
rajendra-desai-ni: Can't we use env variable from `automation_conf.json` or from pipeline env variables instead of hardcoding git user and email here?
Also, it is ideal to use local git config (specific to our repo) instead of setting it globally. Since build machines also build and push (exports of) other components, it might cause problems when you set them globally.

scripts/dev/upstream_merge/automation_conf.json
```diff
@@ -22,6 +22,7 @@
     "toolchain_prefix": "",
     "ssh_target": "",   
     "build_host_ip": "",
+    "build_user": "",
```
rajendra-desai-ni, line 25: Does `build_user` mean the host machine username? If so, how are you going to set this in the pipeline env variable? I mean the script would be running on various available build machines and how are you going to handle it if each machine has a unique username?

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -179,7 +232,22 @@ def create_kernel_pr(args, config, latest_tag, defconfig_changed):
 def run_upstream_merge_script(args):
     original_cwd = os.getcwd()
     print("[INFO] Running upstream RT merge")
+    build_user = (
+        config.build_user
+        if config.build_user
+        else os.getenv("BUILD_USER", getpass.getuser())
```
rajendra-desai-ni: Can't we always use `getpass.getuser()` instead of searching in various places?

rajendra-desai-ni (changes requested):
- [This](https://github.com/ni/nilrt/pull/352/commits/34a8b087122c404ef951e6f4561107628d86f142) commit doesn't have a signed-off tag.
- It is better to squash the 2 commits into one to keep commit history clean.

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Removed the whole part from the script , as it is already taken care of in the pipeline...

scripts/dev/upstream_merge/automation_conf.json
jatinjb444: We are here not providing any value to this , in the script it uses a function called getpass.getuser() , which dynamically takes the current user name

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Sure , will set it to getpass.getuser() only

rajendra-desai-ni (approved):

scripts/dev/upstream_merge/json_config.py
```diff
@@ -50,6 +50,7 @@ def __init__(self, config_path, work_item_id):
         self.build_host_user = kernel_config.get(
             "build_host_user", os.getenv("USER", "builduser"))
         self.pr_enabled = config.get("pr_enabled", False)
+        self.build_user = config.get("build_user")
         self.pr_target_branch = config.get(
```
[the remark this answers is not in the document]
jatinjb444: build_user is now read from the kernel_build section, and a fallback to build_host_user is added to ensure it is always populated and avoids potential None cases.

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -179,7 +233,17 @@ def create_kernel_pr(args, config, latest_tag, defconfig_changed):
 def run_upstream_merge_script(args):
     original_cwd = os.getcwd()
     print("[INFO] Running upstream RT merge")
+    build_user = os.getenv("BUILD_USER", getpass.getuser())
+    print(f"[INFO] Using build user: {build_user}")
 
+    ssh_setup_rc = setup_passwordless_ssh_to_build_machine(
+        config.build_host_ip,
+        build_user
+    )
```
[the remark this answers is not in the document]
jatinjb444: Keeping the current behavior intentionally.

The build_user is derived dynamically using getpass.getuser() to support execution across different build environments without requiring explicit configuration. Since the script may run on different machines/users, relying solely on config.build_host_user can introduce rigidity.

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -91,6 +92,59 @@ def run_cmd(cmd):
...
+
+    if not key:
+        print("[ERROR] Failed to retrieve public key from target")
+        return 1
+
+    # Step 3: Add key to build machine
+    # Step 3: Add key locally (NO SSH needed ✅)
+
+    ssh_dir = os.path.expanduser("~/.ssh")
+    auth_file = os.path.join(ssh_dir, "authorized_keys")
```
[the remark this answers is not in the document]
jatinjb444: Previously the key was always written to the current user's ~/.ssh directory. This has been corrected to use the provided build_user so the key is installed for the appropriate account on the build machine, ensuring proper SSH access.

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -257,8 +333,16 @@ def run_upstream_merge_script(args):
...
     build_root = os.path.dirname(config.kernel_src_dir)
+    tc_result = copy_toolchain_from_nirvana(build_root)
+
+    if len(tc_result) == 3:
+        tc_copy_status, tc_copy_msg, toolchain_dst = tc_result
+    else:
+        tc_copy_status, tc_copy_msg = tc_result
+        toolchain_dst = None
     tc_copy_status, tc_copy_msg, toolchain_dst = copy_toolchain_from_nirvana(
         build_root)
```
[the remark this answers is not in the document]
jatinjb444: The function was being called twice, which caused redundant work and overwrote the initial result. Removed the duplicate invocation and now using the result from a single call consistently.

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -217,10 +281,22 @@ def run_upstream_merge_script(args):
...
 
+    print("[DEBUG] Running safe merge:", cmd)
+
+    rc, out = run_cmd(cmd)
+
+    print("[DEBUG] RC:", rc)
+    print("[DEBUG] OUT:", out)
+
+    result = (rc, out)
+
```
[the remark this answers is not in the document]
jatinjb444: The merge is executed via `bash -c` here because the existing helper (`git_merge`) was not handling this flow correctly in this context and resulted in failures during testing .
Can be done cleanly but here direct  invocation was required .

chaitu236 (approved):
> Add passwordless sudo configuration
Create sudoers entry for required cmds
Avoid manual sudo password prompts
Improve automation for CI execution

Each of these changes should ideally be broken into individual commits. I'm not going to make this a blocker for this time.

PR #357, opened 2026-06-23
docker: assign collision-free container IPs

Shreejit-03:
# Changes

Containers launched on the shared macvlan LAN were colliding on IP addresses. Macvlan IPAM hands out 
the lowest free address from each host's *local* view, so containers started on independent hosts all grabbed .2/.3/.4...,
producing duplicate IPs across the network.

This PR makes container IP assignment collision-free at launch time:

- `docker/nilrt-ctr.sh`: 
  - Add a `run [nilrt|nilrt-slim] [-n N]`  command that scans the LAN for in-use addresses 
  (via `arp-scan` when available, otherwise a parallel ICMP sweep read back from the kernel ARP table), 
picks N verified-free addresses, and launches one container per address. Each address is pinned through
 Compose via the `NILRT_IP` variable. 
  -  Remove the redundant `scale` command, which could not assign distinct static IPs to scaled instances.
- `docker/docker-compose.yml`: Pin `ipv4_address` from `NILRT_IP`
  so each container gets the address chosen by `run`. The variable is
  required to prevent accidental launches without a pinned IP.
- `docker/setup-nilrt-network.sh`: Update the post-setup launch
  hints to point at `nilrt-ctr.sh run` instead of the collision-prone
  bare `docker run --network=nilrt-net`.

# Justification
[AB#3926301](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3926301).

# Testing
Launched containers with `bash docker/nilrt-ctr.sh run nilrt` on a live network . Container came up at a 
scanned-free address instead of .2, confirming collision avoidance.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -4,7 +4,8 @@ x-nilrt-common: &nilrt-common
   tty: true
   restart: unless-stopped
   networks:
-    - nilrt-net
+    nilrt-net:
+      ipv4_address: "${NILRT_IP:?NILRT_IP not set; launch with nilrt-ctr.sh run}"
   labels:
     nilrt.managed: "true"
   healthcheck:
--- a/docker/nilrt-ctr.sh
+++ b/docker/nilrt-ctr.sh
@@ -8,13 +8,13 @@
 #   bash docker/nilrt-ctr.sh <command> [arguments...]
 #
 # Commands:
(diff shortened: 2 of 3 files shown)


Shreejit-03:
Closing this PR as a new [PR](https://github.com/ni/nilrt/pull/358) targeting nilrt/master/scarthgap has been created.

PR #363, opened 2026-06-26
upstream-merge: fix PR automation and logging

jatinjb444:
# Changes

- Standardized PR title and description to align with upstream merge format
- Added handling for cases where no new commits exist (skip duplicate PR creation)
- Improved logging to safely handle missing "Push and PR" entries in merge reports

# Testing

- Successfully ran full automation pipeline end-to-end
- Verified RT merge workflow with latest available tag
- Confirmed behavior when no new commits are present (PR correctly skipped)
- Built kernel locally and verified successful compilation
- Deployed kernel to target system and verified boot
- Rebuilt DKMS modules and confirmed they load correctl

# Process

- Synced local repository with latest upstream branch
- Created isolated working branch for RT merge and PR automation
- Performed RT merge workflow including fetch, merge, and validation
- Ensured clean git history and proper branch usage for PR creation
- Added safeguards to prevent duplicate or empty PR's

__Suggested Reviewers:__
@rajendra-desai-ni 
@ni/rtos

### DIFF
--- a/scripts/dev/upstream_merge/kernel_build_and_test.py
+++ b/scripts/dev/upstream_merge/kernel_build_and_test.py
@@ -14,6 +14,7 @@ from utils.git_commands import (
     git_status,
     git_clone,
     git_checkout,
+    git_pull,
 )
 from utils.git_repo import GitRepo
 from json_config import JsonConfig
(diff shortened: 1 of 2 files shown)


scripts/dev/upstream_merge/kernel_build_and_test.py
rajendra-desai-ni: Can't we use [this](https://github.com/ni/nilrt/blob/431b42a44193cd9693259a86cc1de5473d59cae3/scripts/dev/upstream_merge/utils/git_commands.py#L183) function here instead?

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -298,6 +321,14 @@ def run_upstream_merge_script(args):
 
     git_fetch("origin")
     git_checkout(config.target_branch, create=True, force_checkout=True)
+    
+    git_pull("origin", config.target_branch, rebase=True)
+
+    # ✅ AUTO-CREATE WORKING BRANCH
+    branch_name = f"rt-merge-{config.target_branch.split('/')[-1]}"
+    os.system(f"git checkout -b {branch_name} || git checkout {branch_name}")
```
rajendra-desai-ni, line 333: Can't we use [this](https://github.com/ni/nilrt/blob/431b42a44193cd9693259a86cc1de5473d59cae3/scripts/dev/upstream_merge/utils/git_commands.py#L123) available function for git checkout?

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -241,40 +242,62 @@ def run_rebuild_drivers(config, run_cmd):
...
         local_base_branch=config.target_branch,
     )
+    # Set fork details for GitRepo
+    git_obj.fork_name = config.fork_name
+    git_obj.fork_url = f"https://github.com/{config.username}/linux.git"
 
     status, msg = push_branch_and_create_pr(
         git_obj,
-        branch_name=config.target_branch,
+        branch_name=os.popen("git branch --show-current").read().strip(),
```
rajendra-desai-ni: Can't we use [this](https://github.com/ni/nilrt/blob/431b42a44193cd9693259a86cc1de5473d59cae3/scripts/dev/upstream_merge/utils/git_commands.py#L97) available function?

scripts/dev/upstream_merge/kernel_build_and_test.py
rajendra-desai-ni, line 176: There is no info for status=0 about success now with this change. Don't we want to keep this print info?

scripts/dev/upstream_merge/kernel_build_and_test.py
```diff
@@ -241,40 +242,62 @@ def run_rebuild_drivers(config, run_cmd):
...
     if args.skip_push_and_pr:
         return
+    token = os.getenv("GH_PAT")
+    # ✅ CRITICAL: go to correct repo
+    os.chdir(config.kernel_src_dir)
+    print("[FINAL DEBUG BEFORE PR]")
+    print("username =", config.username)
+    print("fork_name =", config.fork_name)
+    # ✅ CRITICAL: ensure fork remote exists
+    os.system(f"git remote remove {config.fork_name} 2>/dev/null || true")
```
rajendra-desai-ni, line 252: Can't we use [this](https://github.com/ni/nilrt/blob/431b42a44193cd9693259a86cc1de5473d59cae3/scripts/dev/upstream_merge/utils/git_commands.py#L161) available function here instead? Same for below command at line 257.

rajendra-desai-ni (changes requested):
- Signed-off tag has syntax errors
- There are merge conflicts in this PR. We can't merge the PR until conflicts are resolved, please resolve the merge conflicts.

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Updated the merge implementation to use the existing `git_merge` utility. Behavior remains unchanged, and debug handling is preserved.

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: In this case, I’ve intentionally kept the shell-based approach as it provides atomic behavior via `git checkout -b ... || git checkout ...`, which handles the "branch already exists" case in a single step.

The helper `git_checkout` would require additional handling for the failure case when the branch already exists, which could slightly alter the current flow

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Updated to use the `git_branch(show_current=True)` helper to retrieve the current branch name before invoking `push_branch_and_create_pr` .

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: The command `git remote remove <name> 2>/dev/null || true` ensures that the step does not fail if the remote does not already exist . Need to update the helper function which can conflict with the OE-upstream merge (already present code)

scripts/dev/upstream_merge/kernel_build_and_test.py
jatinjb444: Added an explicit informational log when the merge succeeds (rc == 0) to improve visibility, while keeping the existing debug logs and flow unchanged.

rajendra-desai-ni (approved):

PR #366, opened 2026-07-02
upstream_merge: Extract pipeline tool installation

Shreejit-03:
# Changes
Move the tool installation logic from the Azure DevOps pipeline into a reusable install_tools.sh script.

The script installs git-send-email and the GitHub CLI, verifies the downloaded GitHub CLI archive using its published checksum, and prepares the tool directory for the upstream merge workflow.

Update the pipeline to invoke the helper script for both CURRENT and NEXT branch jobs, reducing duplicated logic and making future maintenance easier.

# Justification
[AB#3174387](https://dev.azure.com/ni/94b22d7b-ad7b-4f5e-88f0-867910f91c94/_workitems/edit/3174387).

### DIFF
--- a/scripts/dev/upstream_merge/install_tools.sh
+++ b/scripts/dev/upstream_merge/install_tools.sh
@@ -1,0 +1,90 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+: "${TOOLS_DIR:?TOOLS_DIR environment variable is not set}"
+
+required_cmds=(jq curl tar sha256sum perl mktemp grep realpath)
+for cmd in "${required_cmds[@]}"; do
+    if ! command -v "$cmd" >/dev/null; then
+        echo "Required command '$cmd' is missing on the agent." >&2
+        exit 1
+    fi
+done
+
+# TOOLS_DIR is provided by the caller because the installed binaries must live
+# at a stable path that later, independent pipeline steps add to PATH. Resolve
+# it to an absolute path so the later `cd` into the scratch dir cannot break the
+# copy targets, then create it. Tools are written with fixed names and overwrite
+# in place, so there is no need to wipe the directory first.
+TOOLS_DIR="$(realpath -m -- "$TOOLS_DIR")"
+mkdir -p -- "$TOOLS_DIR"
+
+export PATH="$TOOLS_DIR:$PATH"
+
+echo "Installing git-send-email..."
+
+GIT_SEND_EMAIL_SHA256=15094e3fc8cfbe7931a9aceac99bdc82537d42d3be67ee90518b2371fc7869bf
+
+curl \
+    --fail \
+    --location \
+    --show-error \
+    --connect-timeout 30 \
+    --max-time 300 \
+    https://raw.githubusercontent.com/git/git/v2.44.0/git-send-email.perl \
+    -o "$TOOLS_DIR/git-send-email"
+
+echo "${GIT_SEND_EMAIL_SHA256}  ${TOOLS_DIR}/git-send-email" \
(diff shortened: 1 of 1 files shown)


rajendra-desai-ni (approved):

scripts/dev/upstream_merge/install_tools.sh
```diff
@@ -0,0 +1,111 @@
...
+done
+
+cwd="$(pwd -P)"
+parent="${cwd%/*}"
+[[ -z "$parent" ]] && parent="/"
+
+mkdir -p -- "$TOOLS_DIR"
+TOOLS_DIR="$(cd -P -- "$TOOLS_DIR" && pwd -P)"
+
+if [[ "$TOOLS_DIR" == "/" || "$TOOLS_DIR" == "$cwd" || "$TOOLS_DIR" == "$parent" ]]; then
```
chaitu236: Shouldn't this check happen before the dir is even created in line 19?

scripts/dev/upstream_merge/install_tools.sh
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+: "${TOOLS_DIR:?TOOLS_DIR environment variable is not set}"
```
chaitu236, line 5: Any reason we've to use a user provided TOOLS_DIR? Can't we just create a temp dir using `mktemp -d`? If we do that we can get rid of most of sanity checks below.

scripts/dev/upstream_merge/install_tools.sh
```diff
@@ -0,0 +1,111 @@
...
+    echo "Refusing to use unsafe TOOLS_DIR: $TOOLS_DIR" >&2
+    exit 1
+fi
+
+rm -rf -- "$TOOLS_DIR"
+mkdir -p -- "$TOOLS_DIR"
+
+export PATH="$TOOLS_DIR:$PATH"
+
+echo "Installing git-send-email..."
```
chaitu236, line 25: Why not just use a docker image with all required tools installed? I assume this is run in a pipeline in which case docker will be available.

scripts/dev/upstream_merge/install_tools.sh
```diff
@@ -0,0 +1,111 @@
...
+echo "${GIT_SEND_EMAIL_SHA256}  ${TOOLS_DIR}/git-send-email" \
+    | sha256sum -c -
+
+chmod +x "$TOOLS_DIR/git-send-email"
+
+echo "Installing GitHub CLI..."
+
+ARCH=$(uname -m)
+
+case "$ARCH" in
```
chaitu236: Do we really need this check? Are we expecting this script to be run on any other machine than x86/64? If not, let's remove; no need to complicate the script.

scripts/dev/upstream_merge/install_tools.sh
```diff
@@ -0,0 +1,111 @@
...
+    "$TOOLS_DIR/gh"
+
+chmod +x "$TOOLS_DIR/gh"
+
+command -v git-send-email >/dev/null
+command -v gh >/dev/null
+
+gh --version
+
+echo "All required tools are ready."
```
chaitu236: Image

[nit] Missing newline at end.

chaitu236 (changes requested):

scripts/dev/upstream_merge/install_tools.sh
chaitu236: May not be worth holding up this PR for this.

chaitu236 (approved):

